Issue #7, opened 2024-09-21
 Support more than one instrument


cyclekiller:
Some music such as symphonies has multiple tracks, every row is a track/instrument, for example
![图片](https://github.com/user-attachments/assets/fa40f74d-baf3-4153-8a7f-5275fefe9064)
Could you write some logic that separates each track and handles them one by one?
Currently the logic is closely coupled with single-instrument music (e.g. piano music). Thank you very much!

liebharc:
Hello @cyclekiller ,

Interesting! `homr` was originally designed to convert sheet music for beginners, so this is pushing its limits. Let's break down what's needed to handle your case.

> Currently the logic is closely coupled with single-instrument music (e.g. piano music).

First, for context: `homr` already supports multiple instruments. In fact, piano is treated as two independent voices/instruments (though taking into account that it's one instrument with a broader range could improve detection—another topic entirely). The algorithm detects each staff individually and then identifies braces (brace_dot_detection.py). When it finds staffs grouped by a brace, it combines them. The code calls this "MultiStaff" because I couldn't find a widely accepted term for parallel staff systems. Polyphony can mean different things depending on cultural context https://github.com/NetEase/Polyphonic-TrOMR/issues/3#issuecomment-1829458959. 

Why it doesn’t work for your example: `homr` is designed for camera-captured images and uses filters to remove noise and artifacts. One filter eliminates large vertical black elements, which was added to remove page shadows. Unfortunately, this filter also removes the braces linking your staffs - simply because they are very big. We could adjust this filter (as done in https://github.com/liebharc/homr/tree/issue7bracefilter) to address that.

However, I’m unsure if this would fully solve your problem. `homr` has other limitations which might affect your use case - it only supports treble and bass clefs (G2 and F4 specifically) and lacks support for grace notes, triplets, slurs, and articulations. Adding these features, like additional clef support, could be as simple as training a new model if someone would volunteer on spending a couple of days doing this with their GPU. But that's more of a side note.

The question is: would adjusting the filter to detect the correct number of instruments be enough for your case, despite these limitations? If not, it may be best to keep the code as is since the filter works well in other scenarios. Feel free to experiment with the https://github.com/liebharc/homr/tree/issue7bracefilter branch to see if it meets your needs.

![beethoven-1](https://github.com/user-attachments/assets/b85abefe-490b-47dc-9286-73909c233eae)

cyclekiller:
Okay, I understand where the issue came from. Thank you for your feedback.
[...]

PR #8, opened 2024-09-26
POC to discuss issue #7

liebharc:
This should address #7 by keeping large elements as brackets.

### DIFF
--- a/homr/main.py
+++ b/homr/main.py
@@ -207,7 +207,7 @@ def process_image(  # noqa: PLR0915
         )
         debug.write_threshold_image("brace_dot", brace_dot_img)
         brace_dot = create_rotated_bounding_boxes(
-            brace_dot_img, skip_merging=True, max_size=(100, 1000)
+            brace_dot_img, skip_merging=True, max_size=(100, 2000)
         )
 
         notes = add_notes_to_staffs(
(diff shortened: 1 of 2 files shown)


homr/main.py
```diff
@@ -207,7 +207,7 @@ def process_image(  # noqa: PLR0915
         )
         debug.write_threshold_image("brace_dot", brace_dot_img)
         brace_dot = create_rotated_bounding_boxes(
-            brace_dot_img, skip_merging=True, max_size=(100, 1000)
+            brace_dot_img, skip_merging=True, max_size=(100, 2000)
```
liebharc, line 210: Likely we want to remove the height restriction entirely. Or set it to a huge number to disable it. 2000 just feels very arbitrary, while essentially the goal is to not restrict by height.

homr/xml_generator.py
```diff
@@ -207,6 +207,8 @@ def build_part(  # type: ignore
     for measure in staff.measures:
         part.add_child(build_measure(args, measure, is_first_part, measure_number))
         measure_number += 1
+    if len(staff.measures) == 0:
+        part.add_child(build_measure(args, ResultMeasure([]), is_first_part, measure_number))
```
liebharc, line 211: This actually needs further investigation: Why does the image provided by #7 give empty parts at all?

PR #20, opened 2025-07-20
Add colab notebook, format readme

orangishcat:
- Add a Jupyter notebook that can be run on colab and a link to it in the README
- Format README by enforcing max line length
- When uploading `tabi.jpg` on the README to the colab notebook, the results are pretty inaccurate, as the clef is detected incorrectly. Please let me know if this problem is on my end.
- Remember to change the link to the notebook in the README to your branch if this PR is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 # homr
 
-homr is an Optical Music Recognition (OMR) software designed to transform camera pictures of sheet music into machine-readable MusicXML format. The resulting [MusicXML](https://www.w3.org/2021/06/musicxml40/) files can be further processed using tools such as [musescore](https://musescore.com/).
+homr is an Optical Music Recognition (OMR) software designed to transform camera pictures of sheet music into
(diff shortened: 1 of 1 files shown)


liebharc:
Thanks for the PR @orangishcat ! That's a nice addition and allows people to try out homr more easily.

- The link which must be changed after the merge is the one in the README.md? Is that correct?
- The PR uses MuseScore 3 which is quite old. I once encountered a bug in the musicxml import which was fixed in one of the 4.x versions, but I don't remember details. MuseScore 3 is likely good enough for a demo, and it's easier to install than using the version 4 app images. So, long story short: I think version 3 of MuseScore is fine.
- The issue with tabi was a bug on main. I just pushed a fix. Thanks for spotting this!

liebharc:
Thanks again for the PR @orangishcat 😊! (merged)

orangishcat:
Thanks to you too for reviewing the PR!

PR #23, opened 2025-08-09
Faster imports (easyocr)

aicelen:
Hey,
I noticed homr takes long to load all the packages (up to 6 seconds). One package causing this is `easyocr`. 
I optimized it by only importing the `Reader class `and using `threading `since it's used pretty late (so there's plenty of time to load easyocr). 
This improves the loading time by around 1 second. Using the onnx backend the improvements are even bigger (4 seconds) maybe because torch doesn't need to be loaded.
Can you please merge this before the onnx backend, I think this will make things easier...

Note: The code is not really nice and ruff is throwing a lot of recommendations because of the performance measurement in line 9.

### DIFF
--- a/homr/main.py
+++ b/homr/main.py
@@ -4,9 +4,7 @@ import os
 import sys
 from dataclasses import dataclass
 from threading import Thread
-from time import perf_counter, sleep
 
-t0 = perf_counter()
 import cv2
 import numpy as np
 import segmentation_models_pytorch as smp
@@ -258,8 +256,6 @@ def detect_staffs_in_image(
     ocr = OCR()
     Thread(target=ocr.detect_title, args=(debug, staffs[0])).start()
 
-    t0 = perf_counter()
-
     if len(staffs) == 0:
         raise Exception("No staffs found")
     debug.write_bounding_boxes_alternating_colors("staffs", staffs)
@@ -298,7 +294,7 @@ def detect_staffs_in_image(
 
     debug.write_all_bounding_boxes_alternating_colors(
         "notes", multi_staffs, notes, rests, accidentals
-    )    
+    )
     return multi_staffs, predictions.preprocessed, debug, ocr.get_result
 
 
@@ -401,9 +397,7 @@ def main() -> None:
         parser.print_help()
         sys.exit(1)
     elif os.path.isfile(args.image):
-        t0 = perf_counter()
         process_image(args.image, config, xml_generator_args)
-        print(f"it took {perf_counter() - t0} seconds")
     elif os.path.isdir(args.image):
         image_files = get_all_image_files_in_folder(args.image)
         eprint("Processing", len(image_files), "files:", image_files)
--- a/homr/main.py
+++ b/homr/main.py
@@ -3,7 +3,10 @@ import glob
 import os
 import sys
 from dataclasses import dataclass
+from threading import Thread
+from time import perf_counter
 
+t0 = perf_counter()
 import cv2
 import numpy as np
 import segmentation_models_pytorch as smp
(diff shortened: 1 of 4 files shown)


liebharc:
Hi, I'm still traveling so apologies if answers are brief or if I missed something in your PR.

You can run the whole OCR parallel to the transformer step. So instead of optimizing just the import you could likely spawn a thread, import easyocr on the thread, run the detection and send results back to the main thread. Possibly you would leave easyocr initialized since homr can also run on a folder and process multiple files in one process.

homr/title_detection.py
```diff
@@ -1,32 +1,44 @@
...
 
-import easyocr
+from easyocr import Reader
 
 from homr.debug import Debug
 from homr.model import Staff
 
-reader = easyocr.Reader(["de", "en"], gpu=False, verbose=False)
 
+reader = Reader(["de", "en"], gpu=False, verbose=False)
```
liebharc, line 9: I wonder if we could try a change here:

If we define this global here as "reader: Reader | None = None" and then in detect_title if reader is None, we initialize the reader before we use it. By doing this, all the heavy initialization logic goes onto the thread we spawned and the import in main.py could just become an ordinary import.

main.py
```diff
@@ -0,0 +1,6 @@
+# I added an entry point here.
+# If you don't want this, please remove it.
+
+from homr.main import main
+
+main()
```
liebharc, line 6: I think I once added the same thing and then removed it as  "python homr/main.py" (in a poetry shell) was convenient enough. I'm fine having this, I was just wondering for the reason.

homr/main.py
```diff
@@ -181,7 +183,7 @@ def process_image(  # noqa: PLR0915
         result_staffs = correct_rhythm(result_staffs)
 
         eprint("Writing XML")
-        xml = generate_xml(xml_generator_args, result_staffs, title)
+        xml = generate_xml(xml_generator_args, result_staffs, get_title())
```
liebharc, line 186: We could still log the title here: eprint("Found title: " + title)

homr/main.py
```diff
@@ -41,13 +42,15 @@
...
 from homr.transformer.configs import default_config
 from homr.type_definitions import NDArray
 from homr.xml_generator import XmlGeneratorArguments, generate_xml
 
-os.environ["TF_CPP_MIN_LOG_LEVEL"] = "3"
+def load_easyocr():
+    from homr.title_detection import OCR
 
+# Threaded loading of easyocr for better performance
+Thread(target=load_easyocr).start()
```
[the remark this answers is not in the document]
liebharc: I added another comment in title_detection.py and I think that would be the most straightforward solution: Just let the reader initialize in the thread and then we can use normal imports here.

homr/title_detection.py
```diff
@@ -1,32 +1,44 @@
...
 
-def cleanup_text(text: str) -> str:
+class OCR:
```
liebharc, line 11: Good idea to put this into a class

liebharc (changes requested):

aicelen:
So i put `from easyocr import Reader` into `detect_title()` so it can be cleanly loaded inside one thread. 
The only question left is how we want to handle the situation that `detect_title ()` isn't finished when we run `get_request()`. This situation is very unlikely since it's normally finished between the first and second tromr call.
Currently it would just return an empty string. Of course we could wait for the thread to finish but if the ocr silently fails the homr would also fail.

liebharc:
>  easyocr takes 3 seconds without torch loaded and 1.5 seconds when it's (torch) already loaded

This made me wonder, isn't your overall goal to make homr run without pytorch (during inference)? Otherwise, you still struggle with the large pytorch dependency even after the onnx support in homr (https://github.com/liebharc/homr/pull/22) is added?

So an alternative idea would be to switch to https://github.com/SeldonHZ/TorchfreeEasyOCR , the relevant changes seem to be small and I put them together under https://github.com/liebharc/homr/pull/23 . You might want to check if that also gives you the performance boost you need.

liebharc:
We could also switch to RapidOCR, e.g. https://github.com/liebharc/homr/pull/25 . I don't think the OCR performance is really critical for homr - if the title is wrong then users can easily change that.

Do you have a preference between easyocr (without torch I assume) or RapidOCR?

aicelen:
I don't have a preference. RapidOCR seems to be updated regularly and doesn't have the issue of downloading the model each run.

PR #30, opened 2025-09-18
fix on missing title

eflorent2020:
The fix implemented addresses issue where the OCR function could return None or an empty result, causing the max() function to fail when trying to iterate over ocr_results[0]. The improved null checks ensure the code handles these edge cases gracefully by returning an empty string when no valid OCR results are available.


liebharc:
@eflorent2020 , thanks for the fix!

PR #37, opened 2025-11-23
KV Cache for the Decoder

aicelen:
Hey!
This PR introduces KV Cache for homr's decoder, resulting in significant performance improvements (around 2x faster).

 ### Details on the Implementation
Setting `return_hiddens=True` returns a `LayerIntermediates` object containing the KV cache. Since ONNX can't output  custom objects, we first need to extract the `kv_cache tensors`. This is done in the `ScoreTransformerWrapper`.
- First inference step: Input empty `kv_cache tensors` with `shape=(1,8,0,64)`. 
- Not-first inference step: Rebuild the `LayerIntermediates` object from the `kv_cache tensors` of the prior inference step. Reshaping the context tensor is - in contrast to x_transformers - not done within the onnx model, but in the `ScoreDecoder` because I had problems with onnx treating the if-statements as constants. 
Therefore I slightly modified x_transformer's `AttentionLayers.forward` method.

### Important things to note
- I used `x_transformers==2.4.9` (same as pyproject.toml) instead of `x_transformers==2.7.6` (used in poetry.lock)
- custom_x_transformer.py has been excluded from type_checking (I want to change as little as possible)
- Training has not been tested
- The decoder’s dynamic batch size was removed because it is not used
- This should not influence homr's quality: SER is at 35%
- The new models can be found [here](https://github.com/aicelen/homr/releases/tag/v0.4.0)

### DIFF
--- a/homr/transformer/decoder_inference.py
+++ b/homr/transformer/decoder_inference.py
@@ -65,7 +65,9 @@ class ScoreDecoder:
             x_articulations = out_articulations[:, -1:]
             
             if step != 0: # after the first step we don't pass the full context into the decoder
-                context = kwargs["context"][:, :1] 
+                # x_transformers uses [:, :0] to split the context
+                # which caused a Reshape error when loading the onnx model
+                context = kwargs["context"][:, :1]
 
             inputs = {
                 "rhythms": x_rhythm,
@@ -78,6 +80,7 @@ class ScoreDecoder:
             }
             for i in range(32):
                 inputs[kv_input_names[i]] = cache[i]
+
 
             rhythmsp, pitchsp, liftsp, positionsp, articulationsp, *cache = self.net.run(
                 output_names=[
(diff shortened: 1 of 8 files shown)


aicelen:
Currently Andromr takes around 70s on my (lower-end) phone. My aim is around 30-40s per image. Currently I'm not using the gpu/npu at all which leaves plenty of room for performace improvements. 

I did come up with the custom_x_transformer.py implementation myself (I think I only reworked the if-statements in line 161 to 175). I had problems with onnx making the if-statements constants, so I moved the cache slicing in x_transformers.py line 2494 to ScoreDecoder.generate().

I also think the fixed x_transformers version is problematic. If upgrading x_transformers brings improved quality/inferencetime, you should update it. I would try to implement kv caching for this newer version as well (as I already wrote above it is not that much to modify).

liebharc:
Looks good to me. I will merge this and just add one change on top of it to rename the onnx models. Without the rename, everyone who has already downloaded the models will get an error after the next git pull. The rename triggers the download of the updates models, and this way gets around that error.

PR #39, opened 2025-12-05
Feature/position info

liebharc:
This PR brings the positional information back, which was lost a long time ago with an upgrade of the dependencies.

The positional information of each symbol is derived from the transformer attention weights. The point which receives the most weight is considered to be the position of the symbol. As mentioned in a code comment, that's not very reliable - however it's the closest of what we get from the transformer:

```
        # These coordinates are derived from transformer attention and are inherently imprecise,
        # since the model is optimized for predictive accuracy rather than spatial localization.
        # Because patch tokens are processed in raster order (top-to-bottom, left-to-right),
        # this ordering can be used to reject cases where attention-based coordinates
        # violate monotonic scan constraints and are therefore unreliable.
```

Example image from the debug output: 
tabi_staff-0_output

### DIFF
--- a/training/architecture/transformer/decoder.py
+++ b/training/architecture/transformer/decoder.py
@@ -189,51 +189,44 @@ class ScoreTransformerWrapper(nn.Module):
         performs a power scaling to give more weight to the peaks and then
         calculates the center of mass to get the focus point of the attention.
         """
+
+        # Only use the last 3 layers as the later layers contain the strongest
+        # alignment between attention and semantic object position.
         filtered_intermediate = [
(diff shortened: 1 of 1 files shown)


liebharc:
With the change, the inference on the tabi example image seems to take 25% longer (3.4s per picture instead of 2.7s). 

@aicelen , do you have concerns if we spent some extra time to get the positional information for each symbol? It would allow to work on things like lyrics alignment (https://github.com/liebharc/homr/discussions/38), however it's unclear if someone is really willing to do this.

PR #44, opened 2025-12-23
Fix wrong tuplet rhythm

aicelen:
Tuplets have a wrong rhythm (always too fast), for example in measure 7 of tabi.

I made a PR since I'm unsure how we want to handle the case when n<1. Normally this should not happen, but it could due to the transformer model failing. Is the current handling fine (just returning 1)?

### DIFF
--- a/homr/transformer/vocabulary.py
+++ b/homr/transformer/vocabulary.py
@@ -332,16 +332,11 @@ class SymbolDuration:
         return dur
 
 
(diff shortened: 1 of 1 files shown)


liebharc:
Thanks! I haven't spotted this issue as I looked at the file in MuseScore. Good that it's fixed now. I will merge this and add a unit test on top of it.

PR #47, opened 2025-12-31
Feature/speed up preprocessing

liebharc:
It seems that we can skip some of the preprocessing steps to speed up the processing.

Before, we essentially had three steps:

1. CLAHE on the whole image with a clip limit of 2
2. fastNlMeansDenoising on each individual staff image
3. Another CLAHE on each staff image with a clip limit of 1

My impression is that the denoising mainly removes noise which was amplified by the first CLAHE step. Instead of the three steps, running a single CLAHE on the whole image with a smaller clip limit (also 1) gives the same quality of results.

In addition, this PR removes some cvtColor calls and most of the code now works on grayscale images - with the noticeable exception that the segmentation inference requires a matrix with 3 colors.

### DIFF
--- a/homr/color_adjust.py
+++ b/homr/color_adjust.py
@@ -27,7 +27,7 @@ def get_dominant_color(gray_scale: NDArray, color_range: range, default: int | N
 
 
 def apply_clahe(channel: NDArray) -> NDArray:
-    clahe = cv2.createCLAHE(clipLimit=2.0, tileGridSize=(8, 8))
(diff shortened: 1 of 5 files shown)


aicelen (approved):

liebharc:
Sure, if you share the image with me then I'll take a look.

liebharc:
I've created https://github.com/liebharc/homr/issues/49 for the issue with the accidentals

Issue #45, opened 2025-12-24
Cude13 not supported


BrickDesignerNL:
Follow-up on https://github.com/liebharc/homr/issues/40

**Describe the bug**
```
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2025 NVIDIA Corporation
Built on Fri_Nov__7_19:25:04_Pacific_Standard_Time_2025
Cuda compilation tools, release 13.1, V13.1.80
Build cuda_13.1.r13.1/compiler.36836380_0
```
=>
```
2025-12-24 14:11:09.0226029 [E:onnxruntime:Default, provider_bridge_ort.cc:2251 onnxruntime::TryGetProviderInfo_CUDA] E:\_work\1\s\onnxruntime\core\session\provider_bridge_ort.cc:1844 onnxruntime::ProviderLibrary::Get [ONNXRuntimeError] : 1 : FAIL : Error loading "C:\Users\[username]\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\homr-JdTy6QzD-py3.12\Lib\site-packages\onnxruntime\capi\onnxruntime_providers_cuda.dll" which depends on "cublasLt64_12.dll" which is missing. (Error 126: "Kan opgegeven module niet vinden.")
2025-12-24 14:11:09.0348834 [W:onnxruntime:Default, onnxruntime_pybind_state.cc:1013 onnxruntime::python::CreateExecutionProviderFactoryInstance] Failed to create CUDAExecutionProvider. Require cuDNN 9.* and CUDA 12.*, and the latest MSVC runtime. Please install all dependencies as mentioned in the GPU requirements page (https://onnxruntime.ai/docs/execution-providers/CUDA-ExecutionProvider.html#requirements), make sure they're in the PATH, and that your GPU is supported.

[...]

PR #55, opened 2026-01-16
Faster GPU Acceleration

aicelen:
Hey!
This PR greatly improves GPU performance. The symbol_error_rate benchmark is 3x faster (yes it runs in just 20s for me). 
Before we used to copy all the kv cache tensors to CPU memory to then copy them back to the GPU memory. This took (of course) quite some time.
By using iobinding we can leave the KV cache tensors in GPU memory.

By default the Log Level for OnnxRuntime is 2, I set it to 3 as this hides the Warnings we get when loading the Encoder. I also added an option (--log_level int) which should be used if we encounter an error with OnnxRuntime. 

Performance could be improved even further by doing the detokenization step entirely on GPU (and use batching). But I don't really think it is worth it as would probably take a lot of time.

@liebharc Concerning our recent discussion in #45 about GPU inference not being worth it: I think (with this PR) GPU acceleration is really worth it as it provides great performance improvements over CPU. I would suggest that we add a seperate point in the Readme (saying something like: great performance improvements over GPU; but only for people who are comfortable with using CUDA; limited support by us; please paste the content of `--log_level 2` before opening an issue).


homr/main.py
```diff
@@ -385,6 +385,13 @@ def main() -> None:
         default=GpuSupport.AUTO,
         help=argparse.SUPPRESS,
     )
+
+    parser.add_argument(
```
liebharc: Perhaps we could merge this with the --debug flag. So --debug just sets the necessary log level which would be needed to debug any CUDA issues. This way we keep the CLI interface clean and easy to understand.

homr/segmentation/inference_segnet.py
```diff
@@ -35,16 +37,22 @@ def __init__(self, use_gpu_inference: bool) -> None:
...
         if self.fp16:
-            out = self.model.run(
-                [self.output_name], {self.input_name: input_data.astype(np.float16)}
-            )[0]
+            self.io_binding.bind_cpu_input("input", input_data.astype(np.float16))
         else:
-            out = self.model.run([self.output_name], {self.input_name: input_data})[0]
+            self.io_binding.bind_cpu_input("input", input_data.astype(np.float32))
+
+        self.io_binding.bind_output("output", "cpu")
```
liebharc: Does this provide a speedup as well or is this just make a code cleanup?

liebharc (approved):
Nice! So we were basically IO bound by the KV cache. The changes look good, the only thing I would change is to set the onnx log level based on the already existing --debug flag. 

Just curious, does this also help in your https://github.com/aicelen/Andromr? So do you see an improvement on mobile phones with GPU? And by the way, also feel free to advertise Andromr in the README of homr if you like, so that the project gets more attention.

This is the speed improvement I'm seeing on my desktop (NVIDIA Geforce RTX 4060 Ti + AMD Ryzen 5 3600XT):

```
> time poetry run python homr/main.py tabi.jpg  --gpu no 
...
Result was written to tabi.musicxml

real    0m28.456s

> time poetry run python homr/main.py tabi.jpg  --gpu force
...
Result was written to tabi.musicxml

real    0m17.629s
```

That's a solid 10s saved on tabi. I think that moves the bottleneck from model inference to other sections. Which on the other hand makes me tempted to bring a feature back which older homr versions had: The transformer result was cross-checked against the segmentation result, e.g. do the numbers of clef and notes roughly match. And if that wasn't the case the staff image was augmented and then another attempt was performed. That eliminated some mis-detections. I removed it since it also made everything slower, but since the inference is so fast it could be added again.

The only thing is that I'm still reluctant to offer any CUDA support. For most people I assume it doesn't matter if the processing takes 20s or 30s - they don't have a huge amount of images to be processed. So the hasslefree CPU version is what I would recommend for every user. And for the few power users who benefit from CUDA, I would expect from them that they are willing to invest the time themselves to go through all the troubleshooting resources which are already out there. And if they need support then CUDA and onnx are the better places to get it - and not this repo. Also for things like CUDA13 support, we can offer little as long onnx doesn't support it. Overall I think CUDA support will just cut in our time, with no benefit for any of our projects. I would keep it mentioned in the README, but would go beyond limited support (e.g. no support, but pull requests for fixes and improvements are welcome).

aicelen:
> Just curious, does this also help in your https://github.com/aicelen/Andromr?

No, I'm currently using onnxruntime on cpu. But I already had something similar before: I don't convert the KV cache output Java tensors over to Python.


> Overall I think CUDA support will just cut in our time, with no benefit for any of our projects. I would keep it mentioned in the README, but would go beyond limited support (e.g. no support, but pull requests for fixes and improvements are welcome).

Yes that's true, I'm fine with you solution.


> The transformer result was cross-checked against the segmentation result

Yes I remember trying it out when homr was still using the models by oemer. Does this really improve symbol detection by that much? From my testing the main "problem" are special symbols like slurs and articulations.
If you still decide to use this I would be thankful if you could make it optional to use (maybe even a cli option like --fast) as I'd not use it on mobile.

Also I've been thinking about replacing the segmentation model and the algorithms with something simpler, maybe an Object Detection model? Multistaffs and warped images will probably raise some problems. What do you think about that?

P.S.: I'll add a link to andromr in the readme when it's in the PlayStore. I think that will still take some time though.

PR #56, opened 2026-01-19
Removed scipy as dependency

liebharc:
This PR removes scipy as it's a large dependency which isn't really used that much. The replacement code is LLM generated, which a) has the risk that no one understands it and b) it being incorrect. Based on the tests I ran I have confidence that it's correct. For a) as I haven't found a smaller & credible library which may serve as replacement for scipy, I think it's an acceptable downside. Also, the code seems to be well-structured enough, and it's not at the core of homr, so if the code needs to be replaced again then we can do this.

`find_peaks `gives exactly the same results as the scipy implementation if one compares the results for the same input. 

The dewarping code is harder to check. I doubt that it gives exactly the same results, but the validation result over all files is the same.

For color_adjust I first had some replacement code for the center of mass, but then recognized that a simple clahe gives the same overall performance, so I would vote to just do that.

In terms of execution speed it appears that the code became slightly faster. Which is a bit surprising, but if true then this could also help for Andromr.


aicelen:
Wow that's awesome!
I had problems with packaging scipy into the android app. Therefore I removed staff dewarping and used a custom version for find_peaks.

PR #58, opened 2026-01-27
Repalce Unet++ with Unet

aicelen:
Hey!
I experimented a bit with the Segnet.
I replaced the Unet++ with the simpler Unet architecture. The results seem promising (2x-3x faster; see the results on tabi.jpg below) but I don't have a set of images to test the segnet on. Do you have one?
Have you ever tried using the normal Unet or other more lightweight architectures?
Also it seems that replacing the resnet18 encoder (I tried mobilenetv3_small) is not worth it as it heavily degrades quality while only bringing small performance uplifts.

I marked this WIP as we first need to verify that there aren't any quality degradations.

Old Unet++
unet++_segmentation

New Unet
unet_segmentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ _build/
 *.pth
 *.onnx
 *.pt
+*.onnx.data
 
 # Byprodcuts of a training run
 /seg_unet.keras
@@ -41,6 +42,8 @@ _build/
 /*.musicxml
 /testdata
 /examples
+/canon
+*_try_*
 
 # WSL
 *Zone.Identifier
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented here.
 
 - **Greatly improved GPU inference**.
   Big thanks again to [@aicelen](https://github.com/aicelen) for the contribution!
+- **Faster Segmentation model**
 
 ### Changed
 
--- a/homr/constants.py
+++ b/homr/constants.py
@@ -67,7 +67,7 @@ image_noise_limit = 50
 
 staff_position_tolerance = 50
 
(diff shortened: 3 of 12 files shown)


liebharc:
The grayscale approach seems to give poorer performance during my test runs. The commit prior to that however looks really good. I would add one update to it for "filter_edge_of_vision" which you also wondered about in the past.

If you agree, then I believe that we are close to getting this one merged.

liebharc:
>  Concerning the the transformer, do you think we lack better training data or a better architecture?

That's the hard question. Analog to what we saw of segnet in this PR, I would think that we can get a noticeable performance gain (in terms of accuracy) by spending more time on data augmentation and hyperparameter optimization. The bottleneck for that is likely just GPU availability, so that several parameters can be tried quickly and the results can be compared. Of course someone with more experience in AI will likely also know ways to do these optimizations more quickly.

In my experience homr struggles the most when annotations are added to the sheet music, like chord progressions, fingering tips for violin/flute, or just any marker a musician adds for themselves. One way to deal with that could be another dataset, however I would believe that you could also improve on this by finding good data augmentations.

There is a large community around vision transformers. I don't have a good overview there. The papers I saw suggest that we should be able to replace the resnet to get a performance improvement. However different architectures are also mentioned. Comparing OMR tools is hard as a lot of them have their own specializations (e.g. homr is aimed towards camera pictures) and the lack of an acknowledged benchmark, but my general impression if I read papers about OMR is that homr is achieving the same performance as other systems do, so I think achieving huge improvements is likely not possible without significant effort.

As a bottom line the priority in my mind would be something like this right now: Better data augmentation and hyper parameters, model changes and then last more data.

PR #68, opened 2026-04-12
Graph fusion (MultiHeadAttention

aicelen:
Hi,
this PR introduces custom graph fusion (MultiHeadAttention) to reduce overhead. This brings a speedup of around 10 to 25%. I prototyped the code with Codex and refined it to make it more readable. 
I used name matching, which makes the code cleaner but might break if there are huge changes to transformer. 
If the fusion fails, the model still works like it used to.

I also cleaned up the code for conversion.

And I fixed some bugs concerning CPU fallback in a GPU environment.

@liebharc Can you take a look at this? I would merge it than and upload the fused models.

### DIFF
--- a/training/onnx/convert.py
+++ b/training/onnx/convert.py
@@ -1,10 +1,7 @@
-import os
-
 import torch
 from torch.export import Dim
 
 from homr.segmentation.config import segnet_path_onnx, segnet_path_torch
-from homr.simple_logging import eprint
 from homr.transformer.configs import Config
 from training.architecture.segmentation.model import create_segnet  # type: ignore
 from training.architecture.transformer.decoder import (
--- a/training/onnx/convert.py
+++ b/training/onnx/convert.py
@@ -77,10 +77,6 @@ def convert_encoder() -> str:
 
     path_out = config.filepaths.encoder_path
 
-    if os.path.exists(path_out):
-        eprint(path_out, "is already present")
-        return path_out
-
     # Get Encoder
     model = get_encoder(config)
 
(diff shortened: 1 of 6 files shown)


training/onnx/convert.py
```diff
@@ -77,10 +74,6 @@ def convert_encoder() -> str:
 
     path_out = config.filepaths.encoder_path
 
-    if os.path.exists(path_out):
```
liebharc, line 80: Could we keep these if blocks (this one and the next two)? They allow to just rerun the script e.g. after a transformer training it isn't necessary to convert the segmentation model again.

training/onnx/main.py
```diff
@@ -9,50 +9,49 @@
...
-    path_to_decoder_fp16 = quantization_fp16(path_to_decoder, config.filepaths.decoder_path_fp16)
-    print(path_to_decoder_fp16)
-    path_to_decoder_int8 = quantization_int8(path_to_decoder, config.filepaths.decoder_path)
-    print(path_to_decoder_int8)
+
+    # Conv-heavy models should use static quantization
+    quantization_fp16(path_to_decoder, config.filepaths.decoder_path_fp16)
+    quantization_int8(path_to_decoder)
+
+    fuse_decoder(path_to_decoder)
```
liebharc, line 47: Can we have a fallback here? If quantization or fusing fails then it prints a warning but still gives us the onnx model?

This could be helpful, when we test different transformer architectures and first want to see how the results look like before investing the time to update the runtime optimizations.

liebharc (approved):
Thanks! I left two comments regarding changes which might be helpful. I will also set this to approve already as the comments aren't mandatory.

training/onnx/convert.py
aicelen: I found it more convenient like this, as I had to delete the model before reconverting. Also there's already two separate functions for converting the segnet and the transformer. I'd prefer to remove it but if you want I could introduce it with a flag like --overwrite

liebharc: I usually just run training/onnx/main.py . An "--overwrite" option could be a good compromise. But I'm also really fine if you decide that on your own for now and we see how it works.

aicelen:
Seems like I replaced the models while it was downloading them which caused the run to fail.

PR #70, opened 2026-05-06
Fix docker image build

### LINKED ISSUE #69: Fail to create Docker image
Hi, developers. I'm using homr via docker. [https://github.com/liebharc/homr/blob/main/Dockerfile](url) and got this error
```
/root/.local/bin/poetry run homr -h

Traceback (most recent call last):
...(stacks)...
ImportError: libGL.so.1: cannot open shared object file: No such file or directory
```
so I think we can add `apt install libgl1` in the Dockerfile. And this works for me.

weixlu:
run docker build via [Dockerfile](https://github.com/liebharc/homr/blob/main/Dockerfile) and got this error
```
/root/.local/bin/poetry run homr -h

Traceback (most recent call last):
...(stacks)...
ImportError: libGL.so.1: cannot open shared object file: No such file or directory
```
so I think we can add `apt install libgl1` in the Dockerfile.
This PR Fixes #69

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,8 @@ FROM python:3.11
 
 WORKDIR /app
 
+RUN apt update && apt install -y libgl1
+
 RUN curl -sSL https://install.python-poetry.org | python3 -
 
 RUN git clone https://github.com/liebharc/homr .


liebharc:
@weixlu thanks for fixing this!

PR #71, opened 2026-05-06
Support running on gpu via Docker

weixlu:
When running on gpu in docker, current [Dockerfile](https://github.com/liebharc/homr/blob/main/Dockerfile) lacks `cuda`/`cudnn` env, which `onnxruntime` requires, so I got this msg
```
Failed to load library (...) with error: libcublasLt.so.12: cannot open shared object file: No such file or directory
```
## Solution
1. Since the base image is python:3.11, not anything like nvidia/cuda:12, we need to manually install cuda env
```
RUN /root/.local/bin/poetry run pip install "onnxruntime-gpu[cuda,cudnn]==1.24.1"
```
2. start container like `docker run --rm --gpus all homr`, which is different from the pure-cpu way.

I suggest a new Dockfile.gpu, intended just for running homr on gpu

This PR Refs #69

### DIFF
--- a/Dockerfile.gpu
+++ b/Dockerfile.gpu
@@ -1,0 +1,62 @@
+# Run e.g. with docker build -t homr . && docker run --gpus all --rm -p 8080:8000 homr
+# And then send images: curl -X POST -F "file=@tabi.jpg" http://localhost:8080/process --output tabi.musicxml
+
+FROM python:3.11
+
+WORKDIR /app
+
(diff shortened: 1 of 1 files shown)


aicelen:
Hey,
thank you for the PR. This is the error I also encountered. I think it's a good idea to make a second dockerfile. 

I'd wait until @liebharc took a look at this.

liebharc:
Fine with me, thanks once more for your contribution @weixlu !

Issue #62, opened 2026-02-13
Proper grandstaff detection


liebharc:
`homr` processes sheet music in four stages:

1. detect individual staffs
2. group staffs connected by a brace
3. merge paired staffs into grandstaffs
4. run symbol recognition on the resulting staff images

Step 3 is weak. Piano notation uses a grandstaff (upper + lower staff). Running recognition on each staff separately reduces accuracy because the model loses vertical context across both staffs.

The task: implement a reliable way to decide whether two adjacent staffs form a grandstaff or are independent voices.

Current state: grandstaff creation is done blindly.

Possible direction:

* Use brace detections from the segmentation model.
* Match brace bounding boxes to staff vertical ranges.
* Derive stable pairing rules from brace-staff alignment instead of fixed ordering.

Relevant code:

* `_create_grandstaffs` in `brace_dot_detection.py`
* `create_grandstaffs` in `model.py`

This issue has an example: [https://github.com/liebharc/homr/issues/60](https://github.com/liebharc/homr/issues/60)

fablau:
> I good result would be a somewhat reliable detection if two staffs are a (piano) grandstaff or individual staffs for different types of sheet music: Grandstaff only, grandstaff before another voice, after another voice, or somewhere in between an orchestra piece with many voices.
> 
> I don't have more examples for this. You might find some in the [OpenScore Lieder](https://musescore.com/openscore-lieder-corpus) dataset. But more likely creating a dataset should be the first step for anyone working on this. The algorithm for this would need to work on pictures takes with a camera as this is the goal of the homr project. If camera images aren't available then perhaps [data augmentation](https://github.com/liebharc/homr/blob/feature/next_conv/training/transformer/image_utils.py) can be used.


I see the training package relies on the deepscores dataset as well. That dataset includes many full, orchestral scores. Hasn't the model already been trained on those scores?

weixlu:
@liebharc Hello again :)
I noticed this issue marked as help wanted and you have come up with a good direction. Let me try to implement your plan. 

It seems not quite complex since we can use the brace detection results, not need to retrain.

Issue #60, opened 2026-02-10
Three staves music not recognized?


fablau:
**Describe the bug**
It looks like scores with more than 2 staves are not recognized

**Input Image**
Look at the source image inside the attached .zip file. The .zip file also includes the converted MusicXML file.

[OdeToJoy.zip](https://github.com/user-attachments/files/25217319/OdeToJoy.zip)

**Command You Execute**
poetry run homr OdeToJoy.jpg

aicelen:
Hi!
Homr supports multiple staffs. Until we fixed the issue the easiest workaround would be to use 1abbb36eacb3c6f53d350fed9c35348fca20ab30.

The problem is that the small staffs are not recognized and not passed into the transformer. 
Reverting back to 1abbb36eacb3c6f53d350fed9c35348fca20ab30 fixes the issue. But if we use the new segnet_308 it does not recognize the smaller staffs. Interestingly, replacing segnet_308 with segnet_155 on main does not fix the issue.
Also the image outputs seem to be quite similar between both models. Therefore I'd assume that it should be possible to fix this with some changes to the staff finding code.

Segnet 155:

<img width="970" height="1316" alt="Image" src="https://github.com/user-attachments/assets/18c0bbac-9a17-4631-8344-33039d27d179" />

Segnet 308:

<img width="970" height="1316" alt="Image" src="https://github.com/user-attachments/assets/6cdeb848-05ac-4c10-8e08-2ae4f548ba58" />

fablau:
Ok, great! I'll try that then. Thank you!

liebharc:
> Did the original training dataset include the Deepscore dataset in any way?

Yes for segmentation, so the complete answer is a bit longer. homr uses a two phase approach. The first phase is staff detection, the second transforms each staff image to a text representation of the musical symbols. Deepscore is used to train the segmentation model of the staff detection. The transformer for the second phase is trained on three different datasets (grandstaff, PrIMuS and OpenScores). The transformer datasets contain semantic data, so instead of focusing on individual symbols, they focus on the overall meaning of the image.

fablau:
Ok, finally I found the time to try it, and it works great now. I'd be great if the system could understand more clefs like alto or tenor clefs. Is there any draft or plans to make that work? What else is planned to be implemented here? I'd like to contribute whenever possible.

Thanks again!

liebharc:
Thanks for the confirmation. I will close this issue then.

[...]

PR #72, opened 2026-05-07
Support Grandstaff Detection

weixlu:
This PR implement grandstaff detection method proposed in #62 

- Use brace detections from the segmentation model.
- Match brace bounding boxes to staff vertical ranges.
- Derive stable pairing rules from brace-staff alignment instead of fixed ordering.

## Implementation

In detail, there are 3 steps:

1. we try each two staffs combination, like `(0,1), (1,2), (2,3)`, ...
2. for the current staff pair (like `(0,1)`), we try all braces to see if there is a good match. the pairing rules are:
    i. x_distance: the left side of staff should be close enough (within 5* the average unit size) to the brace symbol
    ii. y_overlap: the top and bottom of the brace symbol should overlap (at least 50% of) the staffs
    iii. final score = y_overlap(higher the better) - x_distance(lower the better)
3. consider some staff may be used more than once(like `([0,1], score=0.5), ([1,2], score=0.8)`), we only select highest score pair, which is `([1,2], score=0.8)`

## Test

- I have run `make test` and `make ci`, seems ok

- Also try `OdeToJoy.jpg` from #60 , and the new grandstaff detection is able to merge staff 2&3 into grandstaff 

former `test_create_grandstaffs_even/odd` are removed

### DIFF
--- a/homr/brace_dot_detection.py
+++ b/homr/brace_dot_detection.py
@@ -131,10 +131,12 @@ def _merge_multi_staff_if_they_share_a_staff(staffs: list[MultiStaff]) -> list[M
     return result
 
 
-def _create_grandstaffs(staffs: list[MultiStaff]) -> list[MultiStaff]:
+def _create_grandstaffs(
+    staffs: list[MultiStaff], brace_dot: list[RotatedBoundingBox]
+) -> list[MultiStaff]:
     if len(staffs) == 0:
         return staffs
-    return [s.create_grandstaffs() for s in staffs]
+    return [s.create_grandstaffs(brace_dot) for s in staffs]
 
 
 def find_braces_brackets_and_grand_staff_lines(
@@ -160,4 +162,4 @@ def find_braces_brackets_and_grand_staff_lines(
         if not any_connected_neighbor:
             result.append(MultiStaff([staff], []))
 
-    return _create_grandstaffs(_merge_multi_staff_if_they_share_a_staff(result))
+    return _create_grandstaffs(_merge_multi_staff_if_they_share_a_staff(result), brace_dot)
--- a/homr/model.py
+++ b/homr/model.py
@@ -419,30 +419,103 @@ class MultiStaff(DebugDrawable):
                 unique_connections.append(connection)
         return MultiStaff(unique_staffs, unique_connections)
 
-    def create_grandstaffs(self) -> "MultiStaff":
-        """
-        Blindly creates grandstaffs, by combining staff 0&1, 2&3, ...
-        for odd numbers of staffs the first staff will be a single staff
-        """
-        n = len(self.staffs)
-        if n == 0:
-            return self
+    """
+    rules to determain whether two staffs should be merged into a grandstaff by current brace:
+    1. x_distance: the left side of staff should be close enough
+       (within 5* the average unit size) to the brace symbol
+    2. y_overlap: the top and bottom of the brace symbol should overlap (at least 50% of) the staffs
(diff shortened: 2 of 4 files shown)


weixlu:
@liebharc Hi, could you help take a look?

homr/model.py
```diff
@@ -419,30 +419,103 @@ def merge(self, other: "MultiStaff") -> "MultiStaff":
...
+                (
+                    self._score_brace_with_staff_pair(symbol, self.staffs[i], self.staffs[i + 1])
+                    for symbol in brace_dot
+                ),
+            )
+            # step2: we try all braces for the current staff pair to see if there is a good match
+            if best_score > 0:
+                pair_scores.append(MultiStaff.GrandStaffPair([i, i + 1], best_score))
+
+        # step3: we can't simply return pair_scores, because some staff may be used more than once,
```
liebharc, line 484: This part could be unit tested, and it might be worth it as I believe it could get more complex in the future.

E.g. we could add some consistency criteria to the weighting which utilizes that notation often has repeated patterns (e.g. one grandstaff for piano over one individual staff for vocals). I don't think we need to add this complexity right now, as it seems to work well as it is. But the tests for the current implementation would be nice.

homr/model.py
```diff
@@ -419,30 +419,103 @@ def merge(self, other: "MultiStaff") -> "MultiStaff":
...
+        # step1: we try each two staffs combination, like (0,1), (1,2), (2,3), ...
+        for i in range(len(self.staffs) - 1):
+            best_score = max(
+                (
+                    self._score_brace_with_staff_pair(symbol, self.staffs[i], self.staffs[i + 1])
+                    for symbol in brace_dot
+                ),
+            )
+            # step2: we try all braces for the current staff pair to see if there is a good match
+            if best_score > 0:
```
liebharc, line 481: Just double-checking if I understood the code correctly: Individual staffs would not get matched this way?

homr/brace_dot_detection.py
```diff
@@ -131,10 +131,12 @@ def _merge_multi_staff_if_they_share_a_staff(staffs: list[MultiStaff]) -> list[M
     return result
 
 
-def _create_grandstaffs(staffs: list[MultiStaff]) -> list[MultiStaff]:
+def _create_grandstaffs(
```
liebharc, line 134: I checked the implementation against https://github.com/liebharc/homr/issues/60 and https://github.com/liebharc/homr/issues/7 and in both cases the algo seems to come to the correct decision

liebharc (commented):
Thanks for looking into this! The code is well-structured, and it made the right decisions for my standard set of test data. I would try to find time for a more detailed review in the next days.

homr/model.py
weixlu: No problem, I have add 3 unit tests for `create_grandstaffs`

- `test_create_grandstaffs`: basic function
- `test_create_grandstaffs_overlapping`: deliberately‌ make 2 braces overlapping, to see if we only pick one
- `test_create_grandstaffs_invalid_brace`: provide an invalid brace and check it'll be ignored

homr/model.py
weixlu: Yes, that's it.

weixlu:
> Thanks for looking into this! The code is well-structured, and it made the right decisions for my standard set of test data. I would try to find time for a more detailed review in the next days.

Thanks for your review and helpful feedback! I’ve addressed the review comments and made the requested changes.
Could you please take another look whenever you have time? No rush at all :)

PR #76, opened 2026-05-15
Add retry for download

weixlu:
This patch add retry for `def download_file()`
1. if some exception raises, we will retry for at most 3 times. exceptions can raise from:
  a. `requests.get(url)` : url inaccessible
  b. `response.iter_content` : download broken half way
2. if still unable to download, we will try a backup url, if exists. At present, `musescore` has a backup url.

### DIFF
--- a/homr/download_utils.py
+++ b/homr/download_utils.py
@@ -8,6 +8,20 @@ from homr.simple_logging import eprint
 
 
 def download_file(url: str, filename: str) -> None:
+    max_tries = 3
+    for i in range(max_tries):
+        try:
+            _download_file(url, filename)
+            return
+        except requests.exceptions.RequestException as e:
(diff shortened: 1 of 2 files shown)


liebharc:
Hi @weixlu , thanks! I see that here you want to be very safe and add both download URLs. With what you said earlier I would however vote to just use the GitHub URL - no backup. It's easier, and I would have some confidence that as it works for you and me, it will work for everyone else as well.

weixlu:
> Hi @weixlu , thanks! I see that here you want to be very safe and add both download URLs. With what you said earlier I would however vote to just use the GitHub URL - no backup. It's easier, and I would have some confidence that as it works for you and me, it will work for everyone else as well.

OK, no problem. I have removed backup url.

PR #77, opened 2026-05-15
Support headless linux in convert_lieder.py

weixlu:
add env for MuseScore, to support running convert_lieder.py on headless Linux

### DIFF
--- a/Training.md
+++ b/Training.md
@@ -17,13 +17,16 @@ Prerequisites:
 - Linux
 - CUDA
 - rsvg-convert, e.g. via `sudo apt install librsvg2-bin`
+- libfuse, e.g. via `sudo apt install libfuse2t64` on Ubuntu24 or `sudo apt install libfuse2` on Ubuntu22/20
+- libjack, e.g. via `sudo apt install libjack-jackd2-0`
 
 Download the datasets and convert them to the format required for training:
 
 - `training/datasets/convert_primus.py`
 - `training/datasets/convert_grandstaff.py`
 - `training/datasets/convert_lieder.py`  
-  This will also download and run MuseScore as an AppImage. If this fails, check your setup to ensure that you can run `datasets/MuseScore`.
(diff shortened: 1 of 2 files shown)


training/datasets/convert_lieder.py
```diff
@@ -171,7 +171,16 @@ def _create_musicxml_and_svg_files() -> None:
 
     eprint("Starting with", len(all_jobs), "jobs, with the first being", all_jobs[0])
 
-    if os.system(MuseScore + " --force -j job.json") != 0:  # noqa: S605
+    env = os.environ.copy()
+    is_headless = not env.get("DISPLAY") and not env.get("WAYLAND_DISPLAY")
+
+    if is_headless:
```
liebharc: We can just always set these env variables. As one would expect, it works independent if it's headless or not and there is no reason we ever should expect the user interacting with the MuseScore UI.

So I would remove the "if" and just add a comment on why this was added

Training.md
```diff
@@ -17,13 +17,16 @@ Prerequisites:
...
+- libjack, e.g. via `sudo apt install libjack-jackd2-0`
 
 Download the datasets and convert them to the format required for training:
 
 - `training/datasets/convert_primus.py`
 - `training/datasets/convert_grandstaff.py`
 - `training/datasets/convert_lieder.py`  
-  This will also download and run MuseScore as an AppImage. If this fails, check your setup to ensure that you can run `datasets/MuseScore`.
+  - This will also download and run MuseScore as an AppImage. If this fails, check your setup to ensure that you can run `datasets/MuseScore`.
+  - Not all files are supported. At the end you'll see something like `Processed 1460/1467 files, skipped 350 files`, which is as expected.
```
liebharc, line 29: Thanks for updating this!

training/datasets/convert_lieder.py
weixlu: OK, I have removed `if`

PR #80, opened 2026-05-15
README - How to run using uvx

Alex23rodriguez:
Easy execution without installing repo using uvx. Note that I don't have a GPU so I cannot actually test if this makes use of it or not

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,14 @@ You might also want to check out [Andromr](https://github.com/aicelen/Andromr), 
 - Poetry
 - Optional: NVidia GPU with CUDA 12.1
 
-## Getting started
+## Getting started (uv)
+
+The easiest way to get started is using `uvx` (`uv` must be installed). However, this might not make use of the GPU
+- `uvx homr <img>`
+- The resulting MusicXML file will be saved in the same directory as the input image
+- To combine the MusicXML results from multiple images, you can use [relieur](https://github.com/papoteur-mga/relieur)
+
+## Getting started (poetry)
 
 - Clone the repository
 - Install dependencies for:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,7 +14,14 @@ You might also want to check out [Andromr](https://github.com/aicelen/Andromr),
 - Poetry
 - Optional: NVidia GPU with CUDA 12.1
 
-## Getting started
+## Getting started (uv)
+
+The easiest way to get started is using `uvx` (`uv` must be installed). However, this might not make use of the GPU
```
liebharc: Hi, thanks for the proposal. The Python ecosystem still feels hard to navigate as there are many options. I'm just not sure if really want to mention all the options in the README as tooling in Python seems to be very subjective.

And strictly speaking the easiest way to get started would be just "pip install homr" as we put releases there?

liebharc: @aicelen , what do you think about this? Should we add the mention of uv?

aicelen: It doesn't work with GPU (i tested it). @Alex23rodriguez Maybe you could remove the "might"?
But I think it's still worth it because:
1. you dont need to use poetry+git
2. it's fast
3. I prefer a CPU only env (if I dont have torch installed) as this doesn't cause any headachs with cuda

Alex23rodriguez: Sorry for the late response. I've updated the file accordingly

Alex23rodriguez: > Hi, thanks for the proposal. The Python ecosystem still feels hard to navigate as there are many options. I'm just not sure if really want to mention all the options in the README as tooling in Python seems to be very subjective.

While it's true that there are many options to navigate the python ecosystem, it has been my experience working with multiple professional teams that uv is quickly becoming the preferred method as it replaces many other tools, though I understand your reservations.
In this case, the advantage of using uv is (apart from the speed) the fact that it can be invoked from anywhere, not just from the location the repo was cloned, since uv is typically installed globally (same as poetry)

liebharc:
Thanks!

PR #83, opened 2026-05-19
Support Multi-GPU Training

weixlu:
This PR enables training transformer on multi-gpu in parrallel. I'm not sure if other developpers need this, so it's totally ok if we don't merge it into main
## Code Change
`HomrTrainer` is subclass from `transformers.Trainer`, so it is already able to train distrubuted. We only need to ensure some work to run barely on rank0, because these works (like saving model, preparing dataset) are serial. 
## Test Run
run by 
```
CUDA_VISIBLE_DEVICES=4,5,6,7 poetry run torchrun 
  --standalone --nnodes=1 --nproc_per_node=4  training/transformer/train.py
```
and I see the 4 gpus running in parrallel
```
+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    4   N/A  N/A         2729517      C   ...mr-jC662fEH-py3.12/bin/python      12650MiB |
|    5   N/A  N/A         2729518      C   ...mr-jC662fEH-py3.12/bin/python      12650MiB |
|    6   N/A  N/A         2729519      C   ...mr-jC662fEH-py3.12/bin/python      12650MiB |
|    7   N/A  N/A         2729520      C   ...mr-jC662fEH-py3.12/bin/python      12650MiB |
+-----------------------------------------------------------------------------------------+
```
and speedup is like 3.4x
- 1 gpu: 1.02it/s
- 4 gpu: 3.40it/s

### DIFF
--- a/Training.md
+++ b/Training.md
@@ -38,6 +38,12 @@ Finally, start the training itself with: `training/train.py transformer`.
 You can check the log and see number of training files to be 197k: `Total number of training files to choose from 197032`
 This takes around 2–4 days.
 
+Distributed Training:
+
+- Distributed training supports single-machine, multi-GPU setups. Using 4 GPUs as an example, run: `CUDA_VISIBLE_DEVICES=0,1,2,3 poetry run torchrun --standalone --nnodes=1 --nproc_per_node=4 training/train.py transformer`
+- Dataset generation and model loading/saving are handled by rank0.
+- It is recommended to use 2 or 4 GPUs to keep training consistent with single-GPU training. For other GPU counts, you need to manually adjust the `gradient_accumulation_steps` parameter. Multi-node, multi-GPU training is theoretically supported, but has not been tested in practice.
+
 ## Results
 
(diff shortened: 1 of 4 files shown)


liebharc:
I'm fine with merging this if it helps you - and envy you a bit for your HW setup ;). I would just need to find the time to confirm that training still works on a single GPU setup before I'd approve this. That will take some days as I'm traveling right now.

weixlu:
Sorry I just find that, the model trained by multi-gpu seems to have worse SER. So I need to do more research before this PR can be merged. Let me check it further...

liebharc:
No worries. It might also just be worth repeating the attempt and see if it was a random event or if this is really related to how things are coordinated between multi GPUs.

weixlu:
I have uploaded the latest code. 

In addition to the previous change (running model checkpoint saving and dataset preparation only on rank 0 in `training/transformer/train.py`), we also need to all-reduce and aggregate loss and metrics across all GPUs (see `training/transformer/metrics.py`).

Now I think this PR is ready for review @liebharc

Issue #73, opened 2026-05-14
Fail to run convert_lieder.py


weixlu:
Hi, I'm running `training/datasets/convert_lieder.py` and see a few issues

### stale download url
This script needs MuseScore ro tun, and download from [url](https://cdn.jsdelivr.net/musescore/v4.2.1/MuseScore-4.2.1.240230938-x86_64.AppImage), which seems to be inaccessible (at least in my region, I'm in Asia and have tried Singapore, Japan and China in my Cloud Server)
So why don't we use the release from github? That's [https://github.com/musescore/MuseScore/releases/download/v4.2.1/MuseScore-4.2.1.240230938-x86_64.AppImage](https://github.com/musescore/MuseScore/releases/download/v4.2.1/MuseScore-4.2.1.240230938-x86_64.AppImage), the same version.

### incompatible with headless Linux
The way we run MuseScore is `os.system(MuseScore + " --force -j job.json")`. It may work on WSL, but on my headless Linux server (deployed in Cloud and connected via remote ssh), I should add `QT_QUICK_BACKEND=software QT_QPA_PLATFORM=offscreen`. 
So I suggest first detect whether it's headless then run MuseScore via
```
mscore_env = os.environ.copy()
mscore_env['QT_QUICK_BACKEND'] = 'software'
mscore_env["QT_QPA_PLATFORM"] = "offscreen"
if subprocess.run(MuseScore + " --force -j job.json", env=mscore_env, shell=True).returncode != 0:  # noqa: S605
    ....
```
Or not change the code but mention this in `Training.md`

### save time
Running `convert_lieder.py` takes quite some time, what about making it parallel? That's easy to do, just call `subprocess.Popen(MuseScore -j job-shared.json)` several times. If that make sense, I can open a PR.
Furthermore, have you thought about uploading converted datasets to [huggingface](https://huggingface.co/datasets)? In this way, other developpers don't have to convert that again. Uploading to huggingface is quite simple, I have done that before.

weixlu:
And after musicxml/svg files are created, I see some other errors
```
INFO: Number of measures ... does not match ...

Error while processing ... Octave change isn't supported

Error while processing ... Backup duration is too long ...

Cannot merge staff at index 1: no next staff available
```
At end I see `Processed 1460/1467 files, skipped 350 files`
Is that as expected?

----------------------
[...]

Issue #81, opened 2026-05-18
Bad slurs&ties


aicelen:
Hi,
the quality of Slurs and Ties is really bad. I'd like to improve it.
I recently found this project: https://github.com/clquwu/Clarity-OMR. It uses a huge model (takes 48gb Vram to train, converting the 3 pages below took 8 minutes on cpu) with similar training material as homr. While it performs worse on Rhythm and notes (it hallucinates like crazy) it performs way better on slurs and ties (while the results are still far from perfect it still is useable unlike homr's detection). 
I think the reason in that lies in it's vocabulary: 
While homr combines the slurs and ties tokens into the articulation branch, Clarity-OMR uses a seperate branch for this with just 8 tokens:
CONNECTION_TOKENS = [
    "tie_start",
    "tie_end",
    "slur_start",
    "slur_end",
    "cresc_start",
    "cresc_end",
    "decresc_start",
    "decresc_end",
]
Do you think it could improve the quality by splitting the current articulation branch into two seperate ones?

The test data I used can be found here: https://imslp.org/wiki/R22_BAN_1_HARM_(Marcoux,_Jean-Fran%C3%A7ois) (it is under [Creative Commons Attribution-ShareAlike 4.0](https://creativecommons.org/licenses/by-sa/4.0/deed.en) so we don't violate copyright).
The .zip file contains the generated musicxml files. I enabled the currently disabled slur detection (music_xml_generator.py line 531-541). 
[test_slurs_result.zip](https://github.com/user-attachments/files/27964621/test_slurs_result.zip)

I have some questions about homr's current implementation: 
- What is the purpose of slurStart_slurStop?
- Why are there slurs between the upper and lower voice (is this a bug in the music_xml_generator)?
- Why aren't slurs and articulations already splitted?
- Have you ever tried to improve the quality of slurs?
- Do we need to differentiate between slurs and ties?

liebharc:
Hi, nice find! I didn't know this model yet. The slur/tie detection in homr was a quick&dirty attempt. At the point where I added the articulations I just was like: Lets see if we can add slurs and ties as well this way. The results weren't good and that is already the end of the story - I never found time to revisit it. That said before you start working on it you might want to take a look at this PR https://github.com/liebharc/homr/pull/65 . It was opened and then closed again, apparently the author didn't want to have it merged. But the work done there looked pretty interesting.

> What is the purpose of slurStart_slurStop?

[...]

PR #84, opened 2026-05-26
fix accuracy for non-positional rhythm

weixlu:
## Describe the Bug
@liebharc mentioned in #73 that sometimes `eval_pitch_accuracy ` stays at 0.6 to 0.7. I reproduce this issue, and see `eval_pitch_accuracy` stays at 0.67, while `eval_rhythm_accuracy` is correnct: 0.99. This doesn't happen very often, but I do reproduce it several times.

## Find Out What's Wrong
I run the smoke test, and manually check transcription result, where I find something interesting. Although eval_pitch_accuracy is 0.67, I see mostly (>90%) the pitches of notes are correct. So why is eval_pitch_accuracy so low?
Below is a function, which shows we can divide rhythm into two catagory: has position and non posiotion.
```
def has_rhythm_symbol_a_position(rhythm: str) -> bool:
    return rhythm.startswith(("note", "rest", "clef"))
```
- for has-position: yes, the pitch accuracy is high, I have checked via smoke test result
- for non-position: no, the pitch accuracy is low. And **this is indeed where the problem is**.

In this PR, I update `metrics.py` and add this record. From a test training, you can see `eval_pitch_nonposRhy_accuracy` is quite low.
```
eval_pitch_accuracy: 0.679374
eval_pitch_nonposRhy_accuracy: 0.089295
eval_pitch_posRhy_accuracy: 0.997130
eval_rhythm_accuracy: 0.996563
....(other metrics)
```
So why is `eval_pitch_nonposRhy_accuracy` so low?
Let's go to `vocabulary.py`. 
- In `def build_rhythm()`, token id 0 is padding. 
- In `def build_pitch()`, token id 0 is nonote. Nonote, e.g. , there isn't a pitch or note, e.g. , this is a non-position rhythm
So the suspect is, for label `nonote`, the training fails. So `eval_pitch_nonposRhy_accuracy` is low.

We can verify this suspect by looking at `class ScoreDecoder`. This line `self.ignore_index = config.pad_token` is wrong, because `pad_token` is token id 0, and `nonote` is also token id 0. If we let `pad_token` to be ignored, then `nonote` is also ignored. As a result, we see the pitch accuracy for non-position rhythm is low.

## How to Fix
That's easy. Don't let `ignore_index = config.pad_token`, but use `mask` to filter the invalid ones. 
After the fix, I have run the training(10 epochs) for 5 times, and don't see eval_pitch_accuracy to be 0.67 anymore.

## What's Next
- I will run training more times to ensure this won't happen again.
- All the trainings are run on #75 , not on branch main. But I assume this won't make a difference.
- To fix this bug, only code change in `decoder.py` is needed. Those in `metrics.py` are only used to diagnostic problems, and it's OK if we remove them.

### DIFF
--- a/training/architecture/transformer/decoder.py
+++ b/training/architecture/transformer/decoder.py
@@ -233,7 +233,9 @@ class ScoreDecoder(nn.Module):
     def __init__(self, transformer: ScoreTransformerWrapper, config: Config):
         super().__init__()
         self.pad_value = (config.pad_token,)
-        self.ignore_index = config.pad_token
+        # NOTE: nonote_token and pad_token currently share the same id (0).
+        # So we cannot use token id as ignore_index, otherwise "nonote" labels are ignored.
+        self.ignore_index = -100
         self.config = config
         self.net = transformer
         self.max_seq_len = config.max_seq_len
@@ -370,6 +372,7 @@ class ScoreDecoder(nn.Module):
 
         if mask.shape[1] == rhythms.shape[1]:
             mask = mask[:, :-1]
+        mask = mask.bool()
 
         # Scheduled Sampling: Two-pass approach
         if self.training and sampling_prob < 1.0:
@@ -436,6 +439,11 @@ class ScoreDecoder(nn.Module):
         loss_consist = beta * self.calConsistencyLoss(
             rhythmsp, pitchsp, liftsp, positionsp, articulationsp, mask
         )
+        rhythmso = self._mask_padding_tokens(rhythmso, mask)
+        pitchso = self._mask_padding_tokens(pitchso, mask)
+        liftso = self._mask_padding_tokens(liftso, mask)
+        articulationso = self._mask_padding_tokens(articulationso, mask)
+        positionso = self._mask_padding_tokens(positionso, mask)
         loss_rhythm = alpha * self.cross_entropy(rhythmsp, rhythmso, label_smoothing=0.1)
         loss_pitch = alpha * self.cross_entropy(pitchsp, pitchso)
         loss_lift = alpha * self.cross_entropy(liftsp, liftso)
@@ -512,6 +520,9 @@ class ScoreDecoder(nn.Module):
             ignore_index=self.ignore_index,
             label_smoothing=label_smoothing,
         )
+
+    def _mask_padding_tokens(self, target: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
+        return target.masked_fill(~mask, self.ignore_index)
 
 
 def init_cache(
--- a/training/transformer/metrics.py
+++ b/training/transformer/metrics.py
@@ -1,9 +1,12 @@
+import os
 from collections import defaultdict
 from typing import Any
 
 import torch
 from torch import Tensor
 from transformers import Trainer
+
+from training.transformer.training_vocabulary import map_rhythm_symbol_with_position
 
 LOSS_COMPONENTS = [
     "loss_rhythm",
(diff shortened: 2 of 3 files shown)


liebharc:
@weixlu thanks once more! That's potentially a really good find. Would you have a fully trained model with this fix, and could you share the pth and trainer state?

homr/transformer/vocabulary.py
```diff
@@ -5,6 +5,8 @@
 from fractions import Fraction
 from typing import Iterable
 
+import torch
```
liebharc: We can't use torch from the homr/ folder as pytorch is an optional dependency. So this code should go to training_vocabulary.py.

weixlu:
> @weixlu thanks once more! That's potentially a really good find. Would you have a fully trained model with this fix, and could you share the pth and trainer state?

[https://get.hidrive.com/8bFSW9cQ](https://get.hidrive.com/8bFSW9cQ) This is a model trained for 10 epochs. If it’s not fully trained yet, please let me know how many more epochs are needed, and I’ll resume training.

homr/transformer/vocabulary.py
weixlu: Ok, I have fixed.

liebharc:
Looks good. Yes, it would be good to see how it performs when training is complete (35 epochs, or early stopping). Thanks!

For this complete run it would be good to not rebase your branches. For homr I always tried to be very traceable with the models. That is why the git commit hash is part of the model name. If we ever find that we went down a wrong road with the model and that a previous model worked much better than it's very easy to see from the model name the exact state of the repo which it was trained on. Also in the past year or so I really tried hard to even commit failed attempts to improve the model and revert them afterward. Because these failures are important learnings and this way it's recorded in the git history what was attempted in the past.

So for your case it would be good to keep your change with https://github.com/liebharc/homr/pull/75 as that's what you are training on.

Thanks again for spotting this issue!

weixlu:
@liebharc OK, I got it. In that case, I'd suggest waiting until #75 is merged before merging this PR.

Regarding #75, is there anything I can do? For example, I'd be happy to run a full training on that branch (with the dataset you send me) and share the model.

liebharc:
#75 mainly needs a training run which gives the same performance as we have on main right now. That's because we have to upload a new model due to the change of vocabulary.

If your change works then we could merge #75 and this one together and upload the training run which had both changes.

weixlu:
> #75 mainly needs a training run which gives the same performance as we have on main right now. That's because we have to upload a new model due to the change of vocabulary.
> 
> If your change works then we could merge #75 and this one together and upload the training run which had both changes.

OK, let me run training on #75 and then share the model

weixlu:
@liebharc Hi, I have finished complete training for both this PR and #75. Smoke test shows SER at 6%-7%. Let me send you `.pth` and `trainer_state.json`.
Somtimes HiDrive is unavailable here in China, so I'll send you a e-mail with the download link from my cloud storage.

liebharc:
@weixlu really thanks a lot for doing this! I've uploaded the model and merged the changes to main. I'm really happy that someone else also looks into the transformer and its training in detail!

weixlu:
> @weixlu really thanks a lot for doing this! I've uploaded the model and merged the changes to main. I'm really happy that someone else also looks into the transformer and its training in detail!

My pleasure :) . I'm really glad the changes were helpful. 

btw, I notice #79 hasn't trained yet. What if I run a full training and share the model again?

liebharc:
Ah I just closed #79. I don't think it's very promising. If you have spare resources then you can give it a go, but it's really not important.

weixlu:
> Ah I just closed #79. I don't think it's very promising. If you have spare resources then you can give it a go, but it's really not important.

OK, then I'll first look into issue #81

PR #86, opened 2026-05-29
Feature/improve slurs

### LINKED ISSUE #81: Bad slurs&ties

aicelen:
This PR aims to improve the models performance on slurs which has been bad in past models.
What I've done so far:
- split the articulation branch (which contained the slur data) into two seperate ones
- Remove tieSlur from rhythm_vocabulary
- remove differentiation between ties and slurs

If anyone reading this has additional ideas to improve the model, please share them :)

I'm currently training the model on a branch based on ddfc63.

@liebharc How important is it to upgrade the branch?


liebharc:
> How important is it to upgrade the branch?

I don't think it's super important. Main has some good improvements, but nothing which should change the performance drastically. So if you just started then it might be worth to take the updates on main and recreate Lieder. If you already did some epochs then I would just let it run.

aicelen:
My training crashed anyways, because I still have some problems with incorrect vocabulary...

liebharc:
> My training crashed anyways, because I still have some problems with incorrect vocabulary...

You can run training/transformer/training_vocabulary.py it will check the token files against the vocabulary and report errors. It's good to do that every time there is a change in vocabulary or dataset to avoid that the training stops after 1 or 2h.

Edit: The script will find issues with the validation dataset which can be ignored.

aicelen:
Is it possible that something is wrong with the code, i get this error:
====== datasets/Lieder-main/flat/lc6209607-1-5.tokens ======
Invalid symbol note_8 G4 _ breathMark_staccato_tenuto _ upper

"breathMark_staccato_tenuto" is also not in the current build_articulation(): https://github.com/liebharc/homr/blob/main/homr/transformer/vocabulary.py

Image:
lc6209607-1-5

Edit: Maybe the vocab is just missing - shall i just add it?

liebharc:
The vocab list is actually generated from the script. So yes, just add anything which is missing.

training/datasets/primus_semantic_parser.py
```diff
@@ -131,29 +131,57 @@ def convert_symbol(cls, symbol: str) -> EncodedSymbol:  # noqa: PLR0911
...
+
 
 def convert_primus_semantic_to_tokens(semantic: str) -> list[EncodedSymbol]:
     symbols = re.split("\\s+", semantic.strip())
     tokens = [PrimusConverter.convert_symbol(sym) for sym in symbols]
     if tokens[-1].rhythm != "barline":
         tokens.append(EncodedSymbol("barline"))
     for symbol in tokens:
-        if has_rhythm_symbol_a_position(symbol.rhythm):
+        if has_rhythm_symbol_a_position(symbol.rhythm): # Where is lower parsed for the position?
```
liebharc: Primus only has single staff images and therefore there is no lower position

Slurs.md
```diff
@@ -0,0 +1,7 @@
+Learnings:
+- Corpus contains slurs which are not used. For example: homr/datasets/Corpus/000051662-1_1_2
```
liebharc: Interesting, I suspect that there are still a lot of issues in the data sets (or how we convert them) which in the end might have an impact on performance

homr/transformer/vocabulary.py
```diff
@@ -663,33 +634,4 @@ def sort_token_chords(
...
-        valid_combinations.append(symbol)
-
-    eprint(
-        "Number of combinations",
-        len(valid_combinations),
-        "- some examples:",
-    )
-    for symbol in random.sample(valid_combinations, 10):
-        eprint(symbol)
+    print(EncodedSymbol("note_4", "F3", empty, empty, empty, "upper"))
```
liebharc: I understand this is a draft PR, just to note: In the end I would like to have the code back which prints the stats

training/datasets/convert_lieder.py
```diff
@@ -8,6 +10,10 @@
 import sys
 from pathlib import Path
 
+script_location = os.path.dirname(os.path.realpath(__file__))
```
liebharc: Why is that needed? Like I get what it does, just not sure why.

aicelen:
I'm at epoch 5 now and the results seem to be not that good:
- the overall accuracy: 0.87
- pitch accuracy: 0.67 (way lower than the other values)

Both of these values are pretty low compared to the training with the Convnext model at the same epoch (overall_accuracy was at 0.98).
Do you have any ideas what could be the reason for that?

Could you share the trainer_state from #84 ?

[trainer_state.json](https://github.com/user-attachments/files/28408521/trainer_state.json)

liebharc:
I checked your code, but didn't see anything wrong (left some minor comments on the way, for later). Did you inspect your data set by looking at a few examples to see if there is anything obvious wrong?

The pitch accuracy hitting a limit at 0.6-0.7 sounds like the issue which was discussed in https://github.com/liebharc/homr/issues/73 but I would have thought that the reason for this was fixed with #84. 

[trainer_state84.json](https://github.com/user-attachments/files/28417776/trainer_state84.json)

Slurs.md
aicelen: I think I should have deleted slurs.md...
I overlooked the tieSlur vocab. I made a script which added slurStart and slurStop to the next note and the note before. So this is not true anymore

training/datasets/convert_lieder.py
aicelen: I think this is some code left over from trying to fix a bug where my changes to convert_lieder didn't apply work (it turned out i ran the wrong homr code). Will remove the code

aicelen:
I think I found the issue: The backbone was remained freezed.
I started training today at Epoch 6 (). As you can see the accuracy jumped to 0.98 in Epoch 7.
The model was notably slower (2.3s/it vs 1.7s/it) compared to yesterday.and more Vram was used (6.5GB vs 4GB). All that while Power Consumption and (Memory) Clock Speed were the same.

The question is why didn't it unfreeze the backbone: Could it be that there's some problems with Pytorch 2.10 (which I use). If I remember correctly it did print "Unfreezing backbone at epoch 2.0" yesterday.

How do we want to continue?
I assume that freezing the backbone for so long will result in a worse model. Therefore I'd suggest to continue training until epoch 10 so we can check if it improves the model performance on slurs&ties. What do you think?

Trainer State for Epoch 7:
[trainer_state.json](https://github.com/user-attachments/files/28419083/trainer_state.json)

Log from current training:
`python3 -m training.train transformer
Total number of training files to choose from 196384
Filtered out 2417 samples with too many ledger lines
Training with 174571 and validating with 19396
Compiling model
Freezing backbone for the first 2 epochs
  0%| | 0/95480 [00:00<?, ?it/s]Unfreezing backbone at epoch 6.0`

liebharc:
Yes I would just let it run for a bit. 10 is enough to see if it works at all. The remaining 25 epochs (up to 35) still seem to be necessary to get the best performance, but for quick checks 10 epochs should be enough.

It's weird that the decoder stays frozen for 6 epochs. epochs_to_freeze is set to 2. I use "2.7.1", I think there was an issue with the newer versions but I don't recall what it was. So not sure what could be the cause of this.

aicelen:
It stayed frozen until I resumed training from the checkpoint of epoch 6

aicelen:
I'm now at epoch 10.
Epoch 10 has a high SER because one image bugged (only output "note_16 . . . . .").
Epoch 9 is at 15%.

Here are the training checkpoints of epoch 9 and 10:
https://get.hidrive.com/3ofbcmvp

I benchmarked the new model (epoch 9 and 10 have identical output) and the existing one (367) on this image:
image

367:
`Writing XML [[clef_G2 _ _ _ upper, keySignature_2 . . . ., note_16 E4 _ _ upper, note_8. F4 # slurStop upper, note_2. D4 _ slurStart upper, barline . . . ., note_4. D4 _ slurStop upper, rest_8 _ _ _ upper, rest_2 _ _ _ upper, barline . . . ., rest_16 _ _ _ upper, note_8. A4 _ slurStart upper, note_8. A4 _ slurStop upper, note_16 D5 _ slurStart upper, note_8 D5 _ slurStop upper, note_8 E5 _ slurStart upper, note_4 E5 _ slurStart_slurStop upper, barline . . . ., newline . . . .]]`


new model:
`Writing XML [[clef_G2 _ _ _ _ upper, keySignature_2 . . . . ., note_16 E4 _ _ _ upper, note_8. F4 # _ _ upper, note_2. D4 _ _ slurStart upper, barline . . . . ., note_4. D4 _ _ slurStop upper, rest_8 _ _ _ _ upper, rest_2 _ _ _ _ upper, barline . . . . ., rest_16 _ _ _ _ upper, note_8. A4 _ _ slurStart upper, note_8. A4 _ _ slurStart upper, note_16 D5 _ _ slurStart upper, note_8 D5 _ _ slurStop upper, note_8 E5 _ _ slurStart upper, note_4 E5 _ _ slurStart_slurStop upper, barline . . . . ., newline . . . . .]]`


Both models have issues with the slurs, while the new one is not fully trained. I observed it in other cases as well that it outputs slurStart for two consecutive notes.  I checked the training data which seems to be fine.
What do you think about those results?

weixlu:
Hi, it looks like there are some issues with the training. I'm happy to pull your branch and run a test on my side to see what's happening.

To address the two points briefly:

1. Pitch accuracy at 0.67: I experienced this too, but it should be resolved on the main branch now.
2. Frozen for 6 epochs: I haven't seen this before. On my side, it only stays frozen for the first 2 epochs.

aicelen:
Hi @weixlu 
The issue with the pitch accuracy at 0.67 was probably caused by the frozen backbone. I'm not sure what caused this though. My current guess is that Pytorch 2.10 is the problem, which version do you use?
Also it seems like the fix in #87 does not apply to slurs, right?
I'm currently on a trip so I can't send you the checkpoints I uploaded to hidrive. 
If you still want to try training, it would be great if you train until something like epoch 8-10 as the model I trained is not that meaningful.

weixlu:
> Hi @weixlu The issue with the pitch accuracy at 0.67 was probably caused by the frozen backbone. I'm not sure what caused this though. My current guess is that Pytorch 2.10 is the problem, which version do you use? Also it seems like the fix in #87 does not apply to slurs, right? I'm currently on a trip so I can't send you the checkpoints I uploaded to hidrive. If you still want to try training, it would be great if you train until something like epoch 8-10 as the model I trained is not that meaningful.

Hi, I'm using torch 2.7.1, as required in `pyproject.toml`. I'm not sure if #87 apply to slurs, let me run training and see

weixlu:
I have done 12 epochs training, which seems fine.

- metrics:
````
{'eval_loss': 0.922996461391449, 'eval_loss_rhythm': 0.8870255002039186, 'eval_loss_pitch': 0.012093984016244903, 'eval_loss_lift': 0.007938883841439065, 'eval_loss_consist': 0.00182746119956761, 'eval_loss_position': 0.0028792133591597187, 'eval_loss_articulations': 0.005183909152860254, 'eval_loss_slurs': 0.006047498999176525, 'eval_rhythm_accuracy': 0.99695693120943, 'eval_pitch_accuracy': 0.9956455255243091, 'eval_lift_accuracy': 0.996892120426536, 'eval_position_accuracy': 0.9990217622456936, 'eval_articulations_accuracy': 0.9985427700533677, 'eval_slurs_accuracy': 0.9981397279972456, 'eval_accuracy': 0.997533139576097, 'eval_runtime': 222.0245, 'eval_samples_per_second': 87.9, 'eval_steps_per_second': 21.975, 'epoch': 12.0}
````

- SER avg: 9%

I also tested the image you mentioned above, it seems some slur stops are missing

```
clef_G2 _ _ _ _ upper
keySignature_2 . . . . .
note_16 E4 _ _ _ upper
note_8. F4 # _ _ upper
note_2. D4 _ _ slurStart upper
barline . . . . .
note_4. D4 _ _ slurStop upper
rest_8 _ _ _ _ upper
rest_2 _ _ _ _ upper
barline . . . . .
rest_16 _ _ _ _ upper
note_8. A4 _ _ slurStart upper
note_16 D5 _ _ slurStart upper
note_8 D5 _ _ _ upper
note_8 E5 _ _ slurStart upper
note_4 E5 _ _ _ upper
barline . . . . .
```

**Comment by weixlu** [comment]
If you need checkpoint, I can upload it later

**Comment by aicelen** [comment]
Thank you for training. Those results are interesting.
The newly trained model misses the second A4 in the second measure and misses more slurs (e.g. slurstart-slurstop in the last note). 
The underlying issue could be errors in the dataset. I already checked the lieder dataset and didn't find an error. It could be that Primus  or grandstaff have errors but both only contain a few slurs. I'll take a look at this later.

Another idea I have is to switch from slurStart and slurStop markers to just a single slured vocabulary.
Example:
Currently we have something like C4 slurStart, A4 _, E4 slurStop
With this approach we would get C4 slured, A4 slured, E4 slured.
It would be harder to reconstruct the output into musicxml and we'd loose some information (slurstart_slurstop) but maybe it is easier for the model as it doesn't need to track if it started/ended a slur.

What do you two think about this (@weixlu  @liebharc)?

**Comment by weixlu** [comment]
I think the plan `C4 slured, A4 slured, E4 slured` is also OK, when a real person reads the measure, it should be like this. But I can't tell if it's better than `C4 slurStart, A4 _, E4 slurStop`.

I haven't carefully read your code, since that's a big change. I'll do it when I have some spare time, and see if I can find anything

**Comment by liebharc** [comment]
I would think that slurstart_slurstop should be detectable in principle as in the image a line ends and another one starts. But I haven't verified if the datasets really always do that. Just brainstorming here: Are we handling chords correctly in the conversion? Like if a slur starts at a chord, which notes receive the slur start?

An in between solution could be this: 

`C4 slurStart, A4 slured, E4 slured C4 slurStart, A4 slured, E4 slured`

You mark slur starts and the slur continuation - the ends can be inferred.

**Comment by aicelen** [comment]
How do we want chords to be handled? Currently the note with the highest pitch is getting the slur marker.

lc4904021-2-2

```
clef_G2 _ _ _ _ upper&clef_F4 _ _ _ _ lower
keySignature_5 . . . . .
note_4 G5 # accent _ upper&note_4 E5 _ _ _ upper&note_8 B4 _ _ _ upper&note_4 E2 _ _ slurStart lower&note_4 E1 _ _ _ lower
note_8 G4 # _ _ upper
note_4 F5 # accent _ upper&note_4 D5 # _ _ upper&note_8 B4 _ _ _ upper&note_4 F2 # _ _ lower&note_4 F1 # _ _ lower
note_8 A4 _ _ _ upper
note_4 E5 _ accent _ upper&note_8 B4 _ _ _ upper&note_4 G2 # _ _ lower&note_4 G1 # _ _ lower
note_8 G4 # _ _ upper
note_4 D5 # accent _ upper&note_8 A4 _ _ _ upper&note_4 B2 _ _ slurStop lower&note_4 B1 _ _ _ lower
note_8 F4 # _ _ upper
barline . . . . .
note_8 F5 # staccato slurStart upper&note_8 D5 # _ _ upper&note_8 A4 _ _ _ upper&note_8 B2 # staccato slurStart lower&note_8 B1 # _ _ lower
note_8 E5 _ staccato _ upper&note_8 C5 # _ _ upper&note_8 G4 # _ _ upper&note_8 C3 # staccato _ lower&note_8 C2 # _ _ lower
note_8 D5 # staccato _ upper&note_8 B4 # _ _ upper&note_8 F4 # _ _ upper&note_8 D3 # staccato _ lower&note_8 D2 # _ _ lower
note_8 C5 # staccato slurStop upper&note_8 E4 _ _ _ upper&note_8 E3 _ staccato slurStop lower&note_8 E2 _ _ _ lower
note_2 B4 _ _ slurStart upper&note_8 G4 # _ _ upper&note_2 E3 # _ _ lower&note_2 E2 # _ _ lower
note_8 D4 _ _ _ upper
note_8 G4 # _ _ upper
note_8 D4 _ _ _ upper
barline . . . . .
note_4 B4 _ _ slurStart_slurStop upper&note_8 G4 # _ _ upper&note_2 B2 _ _ _ lower&note_2 B1 _ _ _ lower
note_8 E4 # _ _ upper
note_4 D5 _ _ _ upper&note_8 B4 _ _ _ upper
note_8 E4 # _ _ upper
note_4 E5 _ _ _ upper&note_8 C5 # _ _ upper&note_2 G2 # _ _ lower&note_2 G1 # _ _ lower
note_8 E4 # _ _ upper
note_8 E5 # _ _ upper&note_8 C5 _ _ _ upper
note_8 E6 # _ slurStop upper&note_8 C6 _ _ _ upper
barline . . . . .
```

**Comment by aicelen** [comment]
Another thing I found is that the Grandstaff dataset has a lot of open slurs. That means that there's a slurStart but there's no slurStop. You can see this in the last A4 of the upper voice.
There're around 38000 slurStart and 34000 slurStop.

There are also some "invisible" ones as well, for example in grandstaff/beethoven/piano-sonatas/sonata28-4/maj2_up_m-21-26
image dataset:
maj2_up_m-21-26

image rendered from kern file:
image

```
clef_G2 _ _ _ _ upper&clef_F4 _ _ _ _ lower
keySignature_5 . . . . .
timeSignature/4 . . . . .
note_4. E5 _ _ _ upper&rest_8 _ _ _ _ upper&note_2 C4 # _ _ lower&note_8 F3 # _ _ lower
note_8 B4 _ _ _ upper&note_8 G3 _ _ _ lower
note_4 A4 # _ _ upper&note_4 F3 # _ _ lower
note_8 F5 # _ _ upper
barline . . . . .
note_8 D5 # _ _ upper&note_4. B4 _ _ _ upper&note_4 B3 _ _ _ lower&note_4. F3 # _ _ lower
note_8 B5 _ _ _ upper
note_8 F5 # _ _ upper&note_8 D4 # _ _ lower
rest_8 _ _ _ _ upper&rest_8 _ _ _ _ upper&note_8 F4 # _ _ lower&note_8 F3 # _ _ lower
barline . . . . .
rest_8 _ _ _ _ upper&note_4. E4 _ _ _ lower&note_8 F3 # _ _ lower
note_4 A5 # _ _ upper&note_4 C5 # _ _ upper&note_8 G3 _ _ _ lower
note_4 F3 # _ _ lower
note_8 A5 # _ _ upper&note_8 C5 # _ _ upper&note_8 F4 # _ _ lower
barline . . . . .
note_8 B5 _ _ _ upper&note_8 B4 _ _ _ upper&note_8 D4 # _ _ lower&note_8 F3 # _ _ lower
clef_G2 _ _ _ _ lower
note_8 F5 # _ _ upper&note_8 D5 # _ _ upper&note_8 B4 _ _ _ lower
note_8 D5 # _ _ upper&note_8 B4 _ _ _ upper&note_8 F4 # _ _ lower
clef_F4 _ _ _ _ lower
note_8 F5 # _ _ upper&note_8 A4 # _ _ upper&note_8 C4 # _ _ lower&note_8 F3 # _ _ lower
barline . . . . .
note_4. E5 _ _ _ upper&note_8 A4 # _ _ upper&note_4 C4 # _ slurStart lower&note_8 F3 # _ _ lower
note_8 B4 _ _ _ upper&note_8 G3 _ _ _ lower
note_4 A4 # _ slurStart upper&note_4 C4 # _ slurStart lower&note_4 F3 # _ _ lower
note_8 F5 # _ _ upper
barline . . . . .
note_4. E5 _ _ slurStop upper&note_8 A4 # _ _ upper&note_4 C4 # _ slurStop lower&note_8 F3 # _ _ lower
note_8 B4 _ _ _ upper&note_8 G3 _ _ _ lower
note_4 A4 # _ slurStart upper&note_4 C4 # _ slurStart lower&note_4 F3 # _ _ lower&rest_4 _ _ _ _ lower
note_8 F5 # _ _ upper
barline . . . . .
```

How would you try to fix that? The slured approach would fix the issue with open slurs but not the problems with errors between the kern and the image file.

**Comment by liebharc** [comment]
> How would you try to fix that? The slured approach would fix the issue with open slurs but not the problems with errors between the kern and the image file.

In the past we tried to identify and remove samples where the ground truth is wrong. In one case we even removed a whole dataset as the accidentals/lifts were unreliable. Can you identify those cases in the grandstaff dataset? The worst case idea: Train without grandstaff and see how well the model does. A model which has never seen the grandstaff dataset during training could in theory then also be used to check the grandstaff dataset for errors and find valid ground truth examples - not sure if it works in practice.

**Comment by aicelen** [comment]
The training is now at epoch 10. The performance on slurs improved slightly compared to weixlu's training run and shows no differences to my first (broken) one (I'm pretty sure this was caused by not using pytorch==2.7.1):

Current:
```
clef_G2 _ _ _ _ upper, keySignature_2 . . . . ., note_16 E4 _ _ _ upper, note_8. F4 # _ _ upper, note_2. D4 _ _ slurStart upper, barline . . . . ., note_4. D4 _ _ slurStop upper, rest_8 _ _ _ _ upper, rest_2 _ _ _ _ upper, barline . . . . ., rest_16 _ _ _ _ upper, note_8. A4 _ _ slurStart upper, note_16 A4 _ _ slurStart upper, note_16 D5 _ _ slurStart upper, note_8 D5 _ _ slurStop upper, note_8 E5 _ _ slurStart upper, note_4 E5 _ _ slurStart_slurStop upper, barline . . . . ., newline . . . . .
```

SER is at 11% (slurs are currently not in the calculation - do they need to be manually added or do we have a script for that?).

[trainer_state.json](https://github.com/user-attachments/files/28766586/trainer_state.json)

You can find the models on my own release page: https://github.com/aicelen/homr/releases/edit/v0.4.0, if we merge this I'd prefer to squash it which would make commit number (417) and hash incorrect.

I also fixed the bug of slurs between the upper and lower voice. In general I'm happy with the current state, here's the result of this branch on tabi:
image

I did not use the latest commits on main, I trained on [Remove flawed files from grandstaff](https://github.com/liebharc/homr/pull/86/commits/2d2d02ed7a40fa41dd8622366eeda1344e5e3340). 

The question is now how we continue:
- The errors in the single-staff test image from above are related to ties, maybe introducing ties could improve the performance. 
- I think the overall result would be good enough to merge it (of course only if I trained the model until epoch 35)

What do you two (@liebharc @weixlu) think?

**Comment by liebharc** [comment]
Hi, sure we can get this merged. Thanks for the improvements!

We have no benchmark for anything but note values and pitches right now. If you rebase this then just make a note in Training.md about the changed commit hash.

Oh and if someone feels up to it the it would be good to find the issue with the newer pytorch version.

**Comment by aicelen** [comment]
Great, shall I restart training with the changes on main or continue from
the already trained epoch 10?

**Comment by weixlu** [comment]
Wow, the transcribed score looks great. I think the issue with ties is fine; the current approach (treating ties as slurs) is a  good solution. It's simple, and intuitive. The accuracy issues might be due to the dataset. For example, the 'open slurs' you found is very important, and when I have some free time, I'll do some filtering on them.

**Comment by weixlu** [comment]
> Currently the note with the highest pitch is getting the slur marker.

Hi, I noticed that this mechanism may have side effect. Take `datasets/grandstaff/beethoven/piano-sonatas/sonata15-4/min3_down_m-120-124.tokens` for example.
```
note_8 B4 _ _ slurStart upper (and other notes)
barline . . . . .
note_4 B4 _ _ slurStart_slurStop upper&note_4. E4 _ _ _ upper (and other notes)
```
Actual sheet music image shows there is a tie ending on `B4` and a new slur starting on `E4`.
image
It seems the `slurStop` on `B4` and `slurStart` on `E4` were merged into a single `slurStart_slurStop`.


Apart from this, sheets like the one below are quite common, so theoretically mark slur on the highest pitch may not be sufficient.
image

liebharc:
I'm not even sure anymore why I added the rule that the marker gets added to the highest pitch. I tried to remember if there was any issue with the dataset or if it made parsing more complex, but we'll can't remember. Usually I added rules like that if there was ambiguity. E.g. chords are sorted by pitch because in notation the order of pitches in a chord doesn't matter, however the transformer being a sequence model just always needs a fixed order. But again here it seems that there should be no ambiguity for slurs as long as the datasets have accurate information. So feel free to change it.

Update: If there are multiple slurs, the matching starts and ends might become a headache. In theory you could add multiple slurStart1, slurStart2, ... but in this case you would need to have a rule how numbers get assigned which the transformer can learn. E.g. go through the sequence symbol by symbol, keep track of the number of open slurs, if new slurs start at the same time then assign new numbers top to bottom.

More thoughts: An algo to normalize the slur numbers can likely be very simple. First you go through the datasets and assign the slur numbers as they appear in the dataset. Music XML has already numbers, grand staff has only ties I believe so you can just use the pitch for defining a number. At this point it's not important what the numbers are as long as the start end pairs match. Then after everything is converted to tokens and all the symbols are sorted you simply go through the sequence in a normalization step. While doing that you simply track a counter of how many slurs have started and replace the slur numbers with the counter value. If the counter exceeds e.g. 5(or 10, 20) you wrap around to 0 to limit the vocab.This numbering of slurs should be easy enough for the transformer to learn.

weixlu:
@liebharc Oh, that's very thorough thinking. Yes, if multiple slurs occru and overlap with each other, things get much more complicated and slurs would need to be numbered. The numbering approach you suggested seems simple and practical.

What I'm more worried about is: since we need to number slurs, can the transformer actually learn that well? My intuition is this: if we're only predicting pitch, that's quite simple—the input is an image of a single note and the output is the pitch. But things like slurs, ties, and tuplets aren't about a single note alone; they're about relationships with each other. I'm not sure whether the current model architechture can learn those kinds of relationships well.

I asked ChatGPT about this, and it suggested adding another attention head after the decoder generates the notes. This  attention head is used to build relationships between notes—including slurs, ties, and tuplets. I think that's an interesting idea, but also ambitious: it would probably be a fairly large change.

aicelen:
I actually tried to distinguish between slurs and ties and also stopped stacking all slurs to the first note in a chord. The results were significantly (it had like 5 slurs less in tabi) worse than what I had before. 
You can find the branch here: https://github.com/aicelen/homr/tree/feature/improve-slurs-ties

liebharc:
In general the decoder is able to learn relationships between notes as the past sequence is part of its context. It already uses that plenty, e.g. to learn to go from notehead positions and clefs to pitch. And it even deals well with clef changes. Even more so, all my attempts to make this easier for the transformer (look for agnostic format in the git history) so far showed that the transformer really has no issue in building this understanding. Based on that, I think it can deal with the slurs as well.

But I don't want to discourage experiments. As mentioned earlier I'm pretty sure that improvements to the model are possible. My experience with ChatGPT ideas was pretty poor so far: They all sound very reasonable and usually never work out. But you might be in the situation that it's much easier for you to test these ideas.

weixlu:
> ChatGPT .... They all sound very reasonable and usually never work out. 

This line got me 😂 — so true.

liebharc:
> I actually tried to distinguish between slurs and ties and also stopped stacking all slurs to the first note in a chord. The results were significantly (it had like 5 slurs less in tabi) worse than what I had before. You can find the branch here: https://github.com/aicelen/homr/tree/feature/improve-slurs-ties

I'm also really fine if we merge this now. Like you obviously put a lot of work into this already and it's a clear improvement!

aicelen:
The model finished training, results look good: It recognizes the ties from the single-staff-test image. The models can be found at https://github.com/aicelen/homr/releases/tag/v0.4.0.

SER is at 7%.

After you reviewed the code I'd squash merge this (I don't think we need all 47 commits) and than add the models and name them correctly after this commit.

liebharc:
The model does a bit worse than what we have currently on main, but I think it's acceptable.

`Total: 6.7 diffs, SER: 5.2%`

The bit poorer performance comes from a lot of small contributions. I looked into them and most of them really look okay. The only weird result is this one:

ohdufroehliche_staff-1_input
(source https://www.franzdorfer.com/o-du-froehliche)

which homr reads as:

grafik

It's almost as it would do an octave shift on the first notes of the top staff. I think we exclude all images with an octave shift from training, but perhaps something is still in there?

But anyway, I don't think that's caused by your changes. So from my side: Feel free to merge :).

PR #87, opened 2026-06-01
fix cross-staff arpeggios

weixlu:
## Describe the Bug
Arpeggios spanning only one staff are incorrectly generated across both staves. The left one is ground truth, while the right one is generated.

image
- 
image

## Root Cause
The root cause is in dataset conversion. Take `lc6810938-1-2` from `datasets/Lieder-main/flat` for example, the arpeggio is in the lower voice.
image
But the `.tokens` shows arpeggiate in both upper and lower voice
```
barline . . . .
note_16 D5 _ arpeggiate upper&note_16 D4 _ _ upper&note_8 B3 b arpeggiate lower&note_8 D3 _ _ lower&note_8 G2 _ _ lower
note_16 D5 _ staccato upper&note_16 D4 _ _ upper
note_24 D5 _ _ upper&note_16 D4 _ _ upper
```
So the model learned from the dataset that, it should extend arpeggio to all staffs.

## Fix
So we need to fix dataset convertion. In `func _fill_in_arpeggiate()`, `arpegiatted_positions.add(symbol.position)` is not enough, because `symbol.position` just marks the beat(timestamp) of the note, while `symbol.symbol.position` is upper/lower voice of the note.
Theoretically this should be enough. I also add a test and it seems ok. But it still needs a retrain (which is in progress) and see if arpeggios are generated correctly.

### DIFF
--- a/training/datasets/music_xml_parser.py
+++ b/training/datasets/music_xml_parser.py
@@ -164,32 +164,35 @@ class TokensMeasure:
             return "upper"
         return "lower"
 
-    def _fill_in_arpeggiate(self, symbols: list[EncodedSymbolWithPos]) -> None:
+    def _fill_in_arpeggiate(self, symbols_with_pos: list[EncodedSymbolWithPos]) -> None:
         """
         In the Lieder dataset we find that an arpeggiate always is rendered
         as it would affect all notes in a chord, even if the MusicXML doesn't
         reflect that.
         """
         arpegiatted_positions = set()
-        for symbol in symbols:
-            if "arpeggiate" in symbol.symbol.articulation:
-                arpegiatted_positions.add(symbol.position)
+        for entry in symbols_with_pos:
+            pos, sym = entry.position, entry.symbol
+            if "arpeggiate" in sym.articulation:
+                # `pos` is the position in the measure, `sym.position` is the upper/lower staff.
+                arpegiatted_positions.add((pos, sym.position))
 
-        for symbol in symbols:
+        for entry in symbols_with_pos:
+            pos, sym = entry.position, entry.symbol
             if (
-                symbol.position in arpegiatted_positions
-                and "arpeggiate" not in symbol.symbol.articulation
-                and symbol.symbol.rhythm.startswith(("note", "rest"))
-                and not symbol.symbol.rhythm.startswith("note_0")
-                and not symbol.symbol.rhythm.startswith("rest_0")
+                (pos, sym.position) in arpegiatted_positions
+                and "arpeggiate" not in sym.articulation
+                and sym.rhythm.startswith(("note", "rest"))
+                and not sym.rhythm.startswith("note_0")
+                and not sym.rhythm.startswith("rest_0")
             ):
-                art = symbol.symbol.articulation
+                art = sym.articulation
                 art_parts = []
                 if art != empty:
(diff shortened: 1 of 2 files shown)


weixlu:
Good. With only 8 epochs, the model is able to keep arpeggios at its onwn staff.
image

training/datasets/music_xml_parser.py
```diff
@@ -173,11 +173,11 @@ def _fill_in_arpeggiate(self, symbols: list[EncodedSymbolWithPos]) -> None:
         arpegiatted_positions = set()
         for symbol in symbols:
             if "arpeggiate" in symbol.symbol.articulation:
-                arpegiatted_positions.add(symbol.position)
+                arpegiatted_positions.add((symbol.position, symbol.symbol.position))
```
liebharc: Can we refactor this to make the naming clearer? I think symbol.position is the position in the measure (so the time) while symbol.symbol.position is the upper/lower staff.

liebharc (commented):
Looks good, I left one minor comment, and then we could merge this to main.

training/datasets/music_xml_parser.py
weixlu: OK, refactor done.

weixlu:
@liebharc My apologies—I just realized some extra code slipped into the merged PR. Specifically, the changes to `get_tuplet_factor()` and `_collect_articulation()` in `training/datasets/music_xml_parser.py` should have been kept for a separate PR.

I still think these changes are worth having, they are intend to better recognize tuplet. Let's discuss if we should keep them, or revert them for now so I can open a clean, dedicated PR for them instead.

Same as above, let's look at the measure `lc6810938`
screenshot_2026-06-02_14-44-42

`datasets/Lieder-main/flat/lc6810938.musicxml` line 2840
```
<notations print-object="no">
          <tuplet type="start" bracket="no"/>
</notations>
<notations>
         slur & staccato
</notations>

... skip some notes

<notations print-object="no">
          <tuplet type="stop"/>
</notations>
<notations>
         slur & staccato
</notations>
```
We can see
1. the note has 2 notations 
2. tuplet is explicitly marked invisible

The current dataset convertion only see the first notation and ignore invisible tuplet, so the generated `lc6810938-1-2.tokens` has 2 bugs:
```
index=0: note_16 D5 _ _ upper&note_16 D4 _ _ upper&note_8 B3 b arpeggiate lower&note_8 D3 _ _ lower&note_8 G2 _ _ lower
index=1: note_16 D5 _ staccato upper&note_16 D4 _ _ upper
index=2: note_24 D5 _ _ upper&note_16 D4 _ _ upper
```
1. note index 0,1,2 are triplet, but assigned with `note_16`, `note_16`, `note_24`
2. note index 0,2 missing staccato

So I make some change to fix these two, but I haven't retrain the model yet. What do you think about it?

liebharc:
Ah sorry I missed that as I hit the merge button. Would be good if we clean up main first, so that we have time to discuss the next PR without pressure.

For invisible tuplets: It was an intentional decision to ignore them. The Lieder dataset contains a lot of them and since they are invisible the model will struggle to find a good rule on when something is a tuplet. As a consequence it overpredicted tuplets during inference. That said we can of course try it again and see how the model behaves these days.

PR #88, opened 2026-06-02
Mac: run segnet on CoreML (MLProgram); keep the transformer on the fastest (CPU fp32) path

GeoffreyPlitt:
## What

homr's only GPU path is hardcoded to `CUDAExecutionProvider`, so on a Mac it always errors and falls back to the CPU. This PR adds ONNX Runtime's **CoreML execution provider** (Apple GPU / Neural Engine) as a first-class GPU backend, so inference runs on the Mac's GPU.

## How

- **New `homr/onnx_providers.py`** centralizes provider selection (`CUDA → CoreML → error`) and reports the IO-binding device. CUDA exposes `"cuda"` device memory that homr binds outputs/KV-cache to; the CoreML EP does not expose user-addressable device memory, so its sessions bind IO on the CPU while the heavy compute runs on the GPU/ANE.
- **`main.py`** treats CoreML as GPU support in the auto-detect gate (previously CUDA-only).
- **`segnet` / `encoder` / `decoder` inference** use the shared selector. The returned device string drives whether outputs and the decoder KV-cache bind to `"cuda"` or `"cpu"`, so the **existing CUDA path is unchanged** and CoreML simply uses CPU-side binding.
- CoreML compute units default to `"ALL"` (GPU + Neural Engine + CPU); set `MLComputeUnits` to `"CPUAndGPU"` to force the GPU only.

## Testing

On macOS 14.4 (Apple Silicon, `onnxruntime==1.26.0`): all three models load on the CoreML EP with **no CPU fallback**, and a sheet-music JPG transcribes end to end to MusicXML. Per-staff TrOmr inference ran ~1.6–3.0 s.

> Draft: opened for discussion / review of the approach. Happy to adjust (e.g. gating behind a `--gpu` value, or the default compute-unit policy).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/homr/main.py
+++ b/homr/main.py
@@ -31,6 +31,7 @@ from homr.model import InputPredictions, MultiStaff
 from homr.music_xml_generator import XmlGeneratorArguments, generate_xml
 from homr.noise_filtering import filter_predictions
 from homr.note_detection import add_notes_to_staffs, combine_noteheads_with_stems
+from homr.onnx_providers import coreml_available, coreml_encoder_enabled, cuda_available
 from homr.resize import resize_image
 from homr.segmentation.config import segnet_path_onnx, segnet_path_onnx_fp16
 from homr.segmentation.inference_segnet import extract
@@ -76,14 +77,14 @@ def get_predictions(
     preprocessed: NDArray,
     img_path: str,
     enable_cache: bool,
-    use_gpu_inference: bool,
+    segnet_use_gpu: bool,
 ) -> InputPredictions:
     result = extract(
         preprocessed,
         img_path,
         step_size=320,
         use_cache=enable_cache,
-        use_gpu_inference=use_gpu_inference,
+        use_gpu_inference=segnet_use_gpu,
     )
     original_image = cv2.resize(original, (result.staff.shape[1], result.staff.shape[0]))
     preprocessed_image = cv2.resize(preprocessed, (result.staff.shape[1], result.staff.shape[0]))
@@ -103,7 +104,7 @@ def replace_extension(path: str, new_extension: str) -> str:
 
 
 def load_and_preprocess_predictions(
-    image_path: str, enable_debug: bool, enable_cache: bool, use_gpu_inference: bool
+    image_path: str, enable_debug: bool, enable_cache: bool, segnet_use_gpu: bool
 ) -> tuple[InputPredictions, Debug]:
     image = cv2.imread(image_path)
     if image is None:
@@ -113,7 +114,7 @@ def load_and_preprocess_predictions(
     image = autocrop(image)
     image = resize_image(image)
     preprocessed = color_adjust.apply_clahe(image)
-    predictions = get_predictions(image, preprocessed, image_path, enable_cache, use_gpu_inference)
+    predictions = get_predictions(image, preprocessed, image_path, enable_cache, segnet_use_gpu)
     debug = Debug(predictions.original, image_path, enable_debug)
     debug.write_image("color_adjust", preprocessed)
 
@@ -157,7 +158,11 @@ class ProcessingConfig:
     write_staff_positions: bool
     read_staff_positions: bool
     selected_staff: int
-    use_gpu_inference: bool
+    # The transformer (encoder/decoder) only benefits from CUDA: its fp16 "GPU"
+    # models are slower than the fp32 ones when they end up on the CPU EP, and
+    # the CoreML EP cannot run the decoder. Segnet additionally supports CoreML.
+    transformer_use_gpu: bool
+    segnet_use_gpu: bool
 
 
 def process_image(
@@ -185,7 +190,7 @@ def process_image(
         debug_cleanup = debug
 
         transformer_config = Config()
-        transformer_config.use_gpu_inference = config.use_gpu_inference
+        transformer_config.use_gpu_inference = config.transformer_use_gpu
 
         result_staffs = parse_staffs(
             debug,
@@ -224,7 +229,7 @@ def detect_staffs_in_image(
     image_path: str, config: ProcessingConfig
 ) -> tuple[list[MultiStaff], NDArray, Debug, Future[str]]:
     predictions, debug = load_and_preprocess_predictions(
-        image_path, config.enable_debug, config.enable_cache, config.use_gpu_inference
+        image_path, config.enable_debug, config.enable_cache, config.segnet_use_gpu
     )
     symbols = predict_symbols(debug, predictions)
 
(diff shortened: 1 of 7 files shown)


liebharc:
Hi @aicelen , I think that falls into your area of expertise - if you have time.

GeoffreyPlitt:
Thanks guys. Full disclosure, I use claude-code to assist with this. I did test it and it *does* use the GPU on Mac, so it works. But if there's any cleanup to do before merging, let me know.

This is really exciting-- I'm a pianist and software engineer, and I love what you've done. Optical Musical Recognition is really cool, and your algorithm / work is truly exceptional.

GeoffreyPlitt:
Und auch, ich spreche deutsch ein bisschen

aicelen:
Hi!
First off thank you for the PR, using CoreML is a good idea.
I'm currently on my phone so I can't really review it properly, I'll take a deeper look at it this weekend. 
I also don't have a Mac, so I can't test it on my own. 
I still have some questions:

1. How much faster is it and which chip are you using? Can you try the different options for MLcomputeUnits? Just use the inference time per staff (maybe the tabi image from the documentation, this is the one I always use).

2. How are loading times of the models? According to the documentation there's a way to use caching to speed it up. It seems quite complicated as we need to manage the cache on our own. Maybe SpecializationStrategy=FastPrediction could improve this.

aicelen:
That was a misclick but I think we can leave it in the ready for review state.

liebharc:
Hi @GeoffreyPlitt , thanks for the PR. Did you have a chance to look into @aicelen questions?

GeoffreyPlitt:
Hi @aicelen — I'm glad you nudged me to run the benchmarks. Admittedly, though I'm a CS-degree software engineer, I'm not an expert in this stuff, so some of the rest here is Claud/Fable run and written...

I ran the benchmarks you asked for (M1, 16 GB, macOS 14.4, ort 1.26.0, the tabi image). Results weren't good...

## Q1: Speed

The PR is actually **~2× slower** than plain CPU, and all `MLComputeUnits` settings perform the same:

| Config | TrOmr per staff |
|---|---|
| CPU baseline (fp32) | **1.21 s** |
| CoreML, any compute-unit setting | 2.2–2.5 s |

ORT's logs show CoreML accepts **zero nodes** of our fp16 models in the default model format — everything silently ran on the CPU anyway, just with the slower fp16 models. My "it's using the GPU" check was wrong: `get_providers()` shows what's *registered*, not what *runs*.

The GPU is real though: with `ModelFormat=MLProgram`, the whole graph compiles and I measured **3.9× on segnet** (1.4 s compile) and **4.8× on the encoder** (but a brutal 60 s compile). The decoder crashes under MLProgram (dynamic KV-cache shape bug in ORT), so the biggest chunk can't use the GPU today.

## Q2: Load times & caching

Default format: all loads ≤0.25 s — nothing compiles, nothing to cache. `FastPrediction` did nothing. Under MLProgram, `ModelCacheDirectory` works and roughly halves loads (60→26 s encoder), and since homr's model filenames are versioned, cache invalidation is automatic — but encoder/decoder loads stay too slow to be a default.

## Proposed rescope

I'm going to work more on this now to do the following (also suggestions from Claude/Fable):

1. **Fix the auto-detect gate** — CoreML-registered must not count as "GPU", since that routes Macs onto the slowest path. As-is the PR shouldn't merge.
2. **Enable CoreML (MLProgram) for segnet only** — 3.9× for a 1.4 s one-time compile. Small, safe, clear win.
3. **Encoder behind an opt-in env var** for batch users; **decoder waits on the ORT fix**.

GeoffreyPlitt:
Rework done. Admittedly, I leaned on Claude/Fable a lot for this.

CoreML is now scoped to where it actually helps:

- **segnet** runs on the GPU via `ModelFormat=MLProgram` — 93/98 nodes placed on CoreML, verified with severity-0 `GetCapability` logs (my earlier `get_providers()` check was misleading: it shows registration, not placement). The compiled model is cached next to the model file; the version-hashed filenames invalidate it automatically.
- **encoder/decoder** stay on the CPU EP with fp32 models. The decoder can't run under MLProgram (ORT 1.26 crashes on the dynamic KV-cache dim — I'll file upstream). The encoder is ~5× faster on the GPU but costs 10–30 s of compile at session creation, so it's opt-in via `HOMR_COREML_ENCODER=1`.
- **CUDA path unchanged** (verified by inspection — no NVIDIA hardware here).

Numbers (M1 16 GB, tabi image, medians of 3 runs):

| | wall | segnet | TrOmr/staff |
|---|---|---|---|
| CPU path (`--gpu no`) | 11.5 s | 3.05 s | 1.207 s |
| previous PR state | 17.2 s | 3.35 s | 2.487 s |
| now (`--gpu auto`) | **9.85 s** | **1.72 s** | **1.17 s** |

Load cost is no longer a concern: segnet's CoreML compile is ~0.4 s warm (disk-cached).

Separate from this PR: batching staffs through the decoder would help all platforms — it's the dominant cost everywhere. I might play with that in a new branch.

aicelen:
Hi, thank you for your effort. 
I took a quick look at the code and also ran it with CUDA: it works and the code looks good so far (there're only some minor things I'd change).
The question is if we still want to merge this: There's only a minor speedup while increasing the complexity of the project. On the other hand it would brings good speedups if it is ran on multiple images (e.g. if we introduce a benchmark). What do you think @liebharc?

liebharc:
The speedups so far look minimal. But the added complexity of the changes also doesn't look big. So I would be fine with merging it.

Batching up images would increase the memory usage, and decrease the processing time. It would likely also make the code much more complex. But you make a good point: As soon as homr is benchmarked (for output quality) on a large dataset you really start caring about processing times. So it likely would be worth it.

homr/onnx_providers.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def coreml_encoder_enabled() -> bool:
+    """Opt-in switch to run the transformer encoder on CoreML (MLProgram).
+
+    The encoder runs ~4.8x faster per forward on the Apple GPU, but compiling
+    the MLProgram takes 26-60 s at session creation, so it only pays off for
+    long batch runs. Off by default.
+    """
+    return os.environ.get("HOMR_COREML_ENCODER", "") not in ("", "0")
```
aicelen: Could you make this a cli argument in main.py? I think it would be useful, if one wants to run homr on a lot of images.
You don't need to make the other env variables cli based though, they seem to be only used for benchmarking.

homr/main.py
```diff
@@ -302,22 +307,26 @@ def get_all_image_files_in_folder(folder: str) -> list[str]:
...
-        ]
-        missing_models = [model for model in models if not os.path.exists(model)]
+        models.extend(
+            [
+                default_config.filepaths.encoder_path,
+                default_config.filepaths.decoder_path,
+            ]
+        )
+    if coreml_encoder_enabled():
+        models.append(default_config.filepaths.encoder_path_fp16)
```
aicelen: If coreml_encoder_enabled() is True and transformer_use_gpu is False, we download both the encoder_path and encoder_path_fp16.

aicelen (changes requested):
There're only two minor things I'd change, apart from that it looks good to me. Thank you :)

GeoffreyPlitt:
Great, here are the updates:

  - --coreml-encoder is now a CLI flag (replaces the HOMR_COREML_ENCODER env var); the other HOMR_COREML_* knobs stay env-only since they're just for
  benchmarking.
  - Fixed the double-download — the encoder and decoder models are now selected independently, so with --coreml-encoder we fetch only the fp16 encoder, not
  both.
  - Fixed a latent crash on the encoder path: the CoreML encoder emits fp16 but the CPU decoder is fp32, so it errored before producing output (that path had
  only been benchmarked at the encoder, never run end-to-end). Context is now cast to the decoder's dtype.
  - Verified on M1: --gpu auto and --gpu auto --coreml-encoder both produce byte-identical output; ruff + mypy clean.

GeoffreyPlitt:
Side note: For context, I'm a jazz pianist/arranger and professional software engineer, so I'm your target audience for this! I've got a goal of eventually leveraging homr to do some AI-directed automatic-blues-lead-sheet arranging. For example, take an old public-domain scan of 1920's blues piano parts, run homr to get musicxml, extract bassline+riff patterns, and then generate a cheat/lead sheet from that. It's a fun project for me because I'm not great at sight-reading staff-notation, but I have a music theory background, and I love lead-sheets.

So suffice it to say, I'm really glad you guys made homr. :)

Batching does sound fun for benchmarking, although I imagine you'd have to re-export the ONNX from the model and reorganize the codebase. Benchmarking itself may be useful, but I think the best part might be identifying output errors either for future work on the model itself, or for adding a "refinement LLM " afterwards that can do little tweaks on minor errors the model makes. For example, I noticed the model sometimes outputs a "wrong octave" for note clusters. Let me know if you want the input sheet-music scans for that test case.

liebharc:
>  Let me know if you want the input sheet-music scans for that test case.

Sure we are always happy to have test images! Even better if you have a cleaned up musicxml file which we can treat as ground truth reference data. I might have found the cause of the octave issue (https://github.com/liebharc/homr/pull/97), but it remains to be seen if that really fixes it. Always feel free to open issues here or drop us a mail.

> I'm your target audience for this!

Just as an anecdote: homr actually was intended as a backend for an app which helps children to learn a music instrument. So it was intended only for basic and rather simple sheet music, but since it was open sourced people put it to more challenging tasks. I'm glad you found a good use for it!

And good luck with your music project! It sounds exciting and Blues is a style I really appreciate!

aicelen (approved):

aicelen:
Great, thank you for the PR @GeoffreyPlitt!

GeoffreyPlitt:
Cool! Glad to meet you. I'll take a look at the octave-fix branch soon, thanks.

PR #89, opened 2026-06-03
Revert tuplet related code in commit "fix cross-staff arpeggios"

weixlu:
#87 mistakenly include code change about tuplet, so we revert them.


liebharc:
Thanks @weixlu !

PR #91, opened 2026-06-05
fix Lieder dataset conversion

weixlu:
https://github.com/liebharc/homr/issues/92


liebharc:
Hi, 

Regarding tremolos: The internal notation doesn't distinguish between ornaments and articulations. I think with your changes tremolos get lost. Or am I missing something?

Apart from that we just need to get the CI pipeline to pass.

weixlu:
> Regarding tremolos: The internal notation doesn't distinguish between ornaments and articulations. I think with your changes tremolos get lost. Or am I missing something?

Oh sorry, I didn't realize the internal notation doesn't distinguish between ornaments and articulations. 
You are right — let me revert the code.

Issue #90, opened 2026-06-03
discussion on invisible tuplets


weixlu:
## Background
Lieder dataset contains a lot of invisible tuplets. 
For example, take a look at measure `lc6810938`
<img width="1133" height="191" alt="screenshot_2026-06-02_14-44-42" src="https://github.com/user-attachments/assets/a6ac2217-ddf3-4223-bb12-1f880e354938" />

`datasets/Lieder-main/flat/lc6810938.musicxml` line 2840
```
<notations print-object="no">
          <tuplet type="start" bracket="no"/>
</notations>
<notations>
         slur & staccato
</notations>

... skip some notes

<notations print-object="no">
          <tuplet type="stop"/>
</notations>
<notations>
         slur & staccato
</notations>
```
`lc6810938-1-2.tokens` has 2 bugs:
```
index=0: note_16 D5 _ _ upper&note_16 D4 _ _ upper&note_8 B3 b arpeggiate lower&note_8 D3 _ _ lower&note_8 G2 _ _ lower
index=1: note_16 D5 _ staccato upper&note_16 D4 _ _ upper
index=2: note_24 D5 _ _ upper&note_16 D4 _ _ upper
```
1. note index 0,1,2 are triplet, but assigned with `note_16`, `note_16`, `note_24`
2. note index 0,2 missing staccato

## Possible Solution
1. Currently tuplets are ignored if they take up > 20%, no matter it is visible or invisible. see `def calc_ratio_of_tuplets` for details
2. @liebharc mentioned that invisible tuplets are designed to be ignored, because it's difficult for the model to recognize it. That makes sense, invisible tuplets are abnormal. Even a real person should feel difficult to recognize.
3. I don't know why `Lieder` has so many invisible tuplets, but since it is just used for training, why don't we change invisible tuplets into visible ones when converting the dataset?

aicelen:
![Scanned-results.jpg](https://github.com/user-attachments/assets/72e3436e-67ce-4e41-b6cf-906dca6c8393)

![input.jpg](https://github.com/user-attachments/assets/5e9a328a-de31-4061-9c13-f394fd6cd8a2)

I assume that the model learns that if there are too many tuplets in one staff, those aren't tuplets. That would explain the behavior in measure 31-35

liebharc:
> I don't know why Lieder has so many invisible tuplets, but since it is just used for training, why don't we change invisible tuplets into visible ones when converting the dataset?

I love the idea. It's simple and could work. 

[...]

PR #93, opened 2026-06-05
redesign add_tuplet_start_stop

weixlu:
#90

### DIFF
--- a/homr/music_xml_generator.py
+++ b/homr/music_xml_generator.py
@@ -8,7 +8,13 @@ import numpy as np
 
 from homr import constants
 from homr.simple_logging import eprint
-from homr.transformer.vocabulary import EncodedSymbol, empty, nonote, sort_token_chords
+from homr.transformer.vocabulary import (
+    EncodedSymbol,
+    SymbolDuration,
+    empty,
+    nonote,
+    sort_token_chords,
+)
 
 
 class ConversionState:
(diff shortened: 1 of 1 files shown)


liebharc:
Let me try to explain how I understand your approach just to double check that I got it right: You split the staff into the lower and upper part. Then you go through it measure by measure. If you find complete tuplets then you extract the tuplets. If the tuplets aren't complete, then you ignore. Is that correct?

weixlu:
> Let me try to explain how I understand your approach just to double check that I got it right: You split the staff into the lower and upper part. Then you go through it measure by measure. If you find complete tuplets then you extract the tuplets. If the tuplets aren't complete, then you ignore. Is that correct?

Yes, that's right. 

The only concern is how to deal with incomplete tuplets. I want to handle it measure by measure, so an incomplete tuplet in one bar won't impact other.

liebharc:
I looked around a bit and tuplets spawning measures seems to be a thing. At least people seem to try to put this into notation SW: https://www.scoringnotes.com/tutorials/create-tuplets-over-a-barline-in-sibelius/ , https://musescore.org/en/node/333221 . Do you have an example with an incomplete tuplet in our dataset? I would just like to check how this looks in the image file to understand if this is an issue in the dataset or really some real world edge case.

Just saw that the PDF from https://github.com/liebharc/homr/pull/94 also has such a case at bar 32&33.

weixlu:
> I looked around a bit and tuplets spawning measures seems to be a thing. 

I also thought about tuplets spanning measures, but since tuplets are already rare, cross-measure ones are even rarer. That's why I didn't plan for this specific case and use a measure-by-measure approach.

weixlu:
> Do you have an example with an incomplete tuplet in our dataset? I would just like to check how this looks in the image file to understand if this is an issue in the dataset or really some real world edge case.

Let me explain what I mean by 'incomplete tuplet'. 

For a 2/4 measure with `note_4 note_12 note_12 note_12`, the second beat is a triplet. If the OMR fails to recognize the last note, it becomes `note_4 note_12 note_12`—an 'incomplete tuplet'. Since the dataset has no errors, these cases won't appear in it.

liebharc (approved):

PR #94, opened 2026-06-08
make tuplet visible

### LINKED ISSUE #90: discussion on invisible tuplets

weixlu:
This is to fix #90 

### Smoke Test
- There is barely tuplet in validation dataset
- SER at 6% after 35 epochs, same as before

### Manual Inspection
- I ran homr on sheet music `divisions for cornet` (from #90), and manually inspected the generated MusicXML. [divisions for cornet.pdf](https://github.com/user-attachments/files/28693739/divisions.for.cornet.pdf)
- Found that most triplets were detected accurately, but not a single sextuplet was detected. This is likely because sextuplets are rare and barely present in the dataset.
- Anyway the result is better than before.

### DIFF
--- a/training/datasets/convert_lieder.py
+++ b/training/datasets/convert_lieder.py
@@ -2,6 +2,7 @@ import json
 import multiprocessing
 import os
 import platform
+import re
 import shutil
 import stat
 import subprocess
@@ -119,8 +120,9 @@ def copy_all_mscx_files(working_dir: str, dest: str) -> None:
 def create_formats(source_file: str, formats: list[str]) -> list[dict[str, str]]:
     jobs: list[dict[str, str]] = []
 
-    # List of files where MuseScore seems to hang up
-    files_with_known_issues = ["sq8940236"]
+    # sq8940236: MuseScore seems to hang up
+    # lc5001945: nested tuplet, not good for training
+    files_with_known_issues = ["sq8940236", "lc5001945"]
     if any(issue in source_file for issue in files_with_known_issues):
         return jobs
     for target_format in formats:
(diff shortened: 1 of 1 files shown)


liebharc:
Thanks! Can you send me the model please to check with my examples?

> Found that most triplets were detected accurately, but not a single sextuplet was detected. This is likely because sextuplets are rare and barely present in the dataset.

True, I'm wondering if we could encode tuplets differently, like the slur changes which are proposed in https://github.com/liebharc/homr/pull/86 : `tupletStart, tuplet, tuplet, tuplet, tuplet, tuplet_` would then be a sextuplet. But not even sure if that works.

weixlu:
> Thanks! Can you send me the model please to check with my examples?

Sure, I have send you via email. This time the training is run on multi-gpu, so if we see any regression, it may be caused by either this PR or the multi-gpu training.

> True, I'm wondering if we could encode tuplets differently, like the slur changes which are proposed in #86 : `tupletStart, tuplet, tuplet, tuplet, tuplet, tuplet_` would then be a sextuplet. But not even sure if that works.

What a coincidence! After reading about the slur change, I also had the idea to do something similar on tuplet. But this may introduce fairly significant change, so I'd suggest we first evaluate the impact of #86 and see if the results look promising

liebharc:
> it may be caused by either this PR or the multi-gpu training.

The results looked good! I will merge this. Sounds like we can also get https://github.com/liebharc/homr/pull/83 ready for a merge.

PR #97, opened 2026-06-13
Octave shifts must removed, as we do already for octave changes

liebharc:
MusicXML has both octave change and octave shifts. We didn't filter out the shifts yet, which likely causes https://github.com/liebharc/homr/pull/86#issuecomment-4698210995 . Note that the lieder dataset must be recreated and a new training runs needs to happen for the fix to work.


liebharc:
Change was squashed to main

GeoffreyPlitt:
@liebharc Since this was a training-data preprocessing fix, will you be updating the model soon?

liebharc:
Not sure about the timeline. There is a lot of movement right now in the training pipeline, as there are a couple of improvements we want to try. It will likely take a bit for things to finish and then a couple of runs to fix bugs.

GeoffreyPlitt:
Awesome! I have some training data in the form of some complicated multi-page sheet music (95% success by my eye, when I examine the output) in case you need some.

liebharc:
Sure, diverse training data helps us a lot. Do you also have ground truth data with it? So the 100% correct music XML? And can we legally share it, since as a community project we would need to publish our training data?

GeoffreyPlitt:
Darn, the answer to both is no, I should have thought of that. But I've seen tons of musicxml on https://musescore.com/, much in the public domain I believe, although I'm sure you've seen it already.

liebharc:
No worries. And yes, we have a lot of musescore files in various complexity. Plus we just got some more, that's one of the changes which might make it into the training pipeline.

PR #98, opened 2026-06-13
Fix, sometimes grandstaffs weren't paired

liebharc:
Hi @weixlu , 

I have one case were the grandstaff detection doesn't work. It's two grandstaffs, but it only pairs the second one, so it runs the transformer with two individual staffs and a grandstaff. 

I'm just not sure if my fix makes sense :D. So I'd be happy to have your opinion. From my understanding y_overlap is a distance and so the abs part would be missing. I can also send you the image by mail if you want to take a look yourself. 

The fix would improve the benchmark results of https://github.com/liebharc/homr/pull/86#issuecomment-4698210995to `5.4 diffs, SER: 4.9%`

### DIFF
--- a/homr/model.py
+++ b/homr/model.py
@@ -440,7 +440,7 @@ class MultiStaff(DebugDrawable):
 
         symbol_min_y = symbol.center[1] - symbol.size[1] / 2
         symbol_max_y = symbol.center[1] + symbol.size[1] / 2
-        y_overlap = min(symbol_max_y, lower_staff.max_y) - max(symbol_min_y, upper_staff.min_y)
+        y_overlap = abs(min(symbol_max_y, lower_staff.max_y) - max(symbol_min_y, upper_staff.min_y))
 
         if (
             x_distance < x_distance_threshold


weixlu:
Looks like your fix is working (based on the benchmark result), but I'm not sure how it did this.
I have draw a picture, where you can see, theoretically, `min(symbol_max_y, lower_staff.max_y)` should be bigger than `max(symbol_min_y, upper_staff.min_y)`
image

Do you mind sharing that image? I'd love to take a closer look.

weixlu:
It may sound a little bit funny, but, I tried to "draw" a brace for the first grandstaff, and (without any code change) it works just fine.
yankee2
image

I also add some debug logs in grandstaff detection. It shows that the second grandstaff can be detected, while the first one can not, unless I draw a brace.

liebharc:
@weixlu actually I really like the creativity! I think I see your point and it shows clearly that my fix isn't the right direction. The previous algo also used the barlines as information if two staffs form a grandstaff, the new algo only relies on the brace. In this example that fails. However, it's debatable if we want to support this case. Cut off images are just hard.

weixlu:
Yes, I agree that transcribing incomplete sheet music is very difficult. To keep things simple, I think it's reasonable to skip this case.

PR #100, opened 2026-06-16
Feature/benchmark

liebharc:
**Disclaimer:** This task was almost entirely completed by Claude Code.

**What it does:**
- Run a range of OMR tools: *homr*, *oemer*, *smt*, and *transcoda*
- Against two datasets: *smb* and *polish-scores*
- Produces a SQLite database file containing all run information: raw predicted and expected outputs, plus statistics. Files from the most interesting runs are available here: [benchmark release](https://github.com/liebharc/homr/releases/tag/benchmark)

**Parser robustness**

Benchmarking is hard. A particular challenge is that meaningful results require robust parsing across different data formats. To address this, significant effort went into parsing:

- Two parsing options are supported for both Humdrum kern and MusicXML:
  - **native** - the parsers already used in homr for training runs
  - **music21** - new parsers based on the music21 library
- Several utilities are included to cross-check the parsers against each other
- music21 and hum2xml can also be run within the benchmark itself; in that case the benchmark measures conversion errors rather than OMR quality
- After thorough investigation and review of examples, the homr **native** parsers appear to be the better choice
- Further details, including the risks of using the same parsers for both training and benchmarking, are documented in `Benchmark.md`

**Fairness & transparency**

The benchmark should be fair overall. Publishing the [SQLite files](https://github.com/liebharc/homr/releases/tag/benchmark) is intended to lower the barrier for anyone who wants to verify the results themselves - and hopefully encourage contributions toward improved benchmarks.

**Known limitations of homr**

Two areas where homr currently underperforms:

- **Triplet detection:** Training excludes triplets without visible triplet markings. Polish scores contain many such cases, causing homr to miss most of them.
- **Octave changes:** Training also excludes octave changes. Polish scores contain these as well, leading to failures in detecting them.

Simply adding unmarked triplets and octave changes to training may not be sufficient - they were excluded in the first place because including them degraded detection quality elsewhere.


liebharc:
The benchmark report for homr as html (exported from the SQLite files):
[smb_homr.html](https://github.com/user-attachments/files/29108133/smb_homr.html)
[polish-scores_homr.html](https://github.com/user-attachments/files/29108134/polish-scores_homr.html)

weixlu:
Happy to see this benchmark! Now we can clearly see which parts are transcribed incorrectly, instead of just a SER number.

I have run your code, and see 2 minor issues:
1. in `validation/show_diff.py`, it says the usage is `--sample-id ID`, but in `Benchmark.md`, it's `--sample 3` 
2. This patch introduces new dependency: library `datasets`, but this hasn't been added to `pyproject.toml` yet. Not sure what's the version you're using. I'm using datasets=5.0.0, and it seems fine.

weixlu:
I'm currently running `polish-scores`, and yes I got a similar result as yours. 
```
Processed 112 samples: 110 succeeded, 2 failed.

Aggregate Statistics:
  Overall NED :
    Mean:      24.85%
    Median:    19.71%
    Min:        4.72%
    Max:       62.99%
    Variance: 238.2348%²
```
I can't run SMB for now, because it seems to require authentication.

liebharc:
Thanks for taking a look at the changes @weixlu ! I updated the PR to address your points, including an instruction on how to get access to the SMB dataset.

PR #102, opened 2026-06-18
fix svg&xml pages mismatch

weixlu:
In `convert_lieder.py`, there is a `if` branch for `len(svg_files)` and `len(xml_pages)` mismatch. This PR is intend to fix this mismatch, which has 2 root causes.
```
def convert_xml_and_svg_file():
        pages = split_into_pages(voices)
        svg_files = get_position_from_multiple_svg_files(str(file))
       
        if len(pages) != len(svg_files):
                ......
```
## 1. Blank SVG Page
Sometimes MuseScore renders music-less svg pages. For example, Lieder-main/flat/lc6611874-4.svg is basically blank. So I add this filter.
image
```
def get_position_from_multiple_svg_files:
        music_file = get_position_information_from_svg(svg)
        if music_file.number_of_measures != 0:
            result.append(music_file)
```
## 2. forward-repeat barline shouldn't be counted.
image

(datasets/Lieder-main/flat/lc4985931-1.svg) This forward-repeat barline is counted in svg but not in xml pages, which introducing `total_xml_measures < total_svg_measures` in `convert_lieder.py:Line 476`
To fix this, I add a new class `BarLine` which distinguish thin and heavy barline. If a barline is heavy and leftmost and close enough to the left side of the staff, it will not be counted(see `def add_bar_line()`).
## Test
I havn't retrained yet. Only run `convert_lieder.py` and it seems OK

### DIFF
--- a/training/datasets/convert_lieder.py
+++ b/training/datasets/convert_lieder.py
@@ -470,25 +470,8 @@ def convert_xml_and_svg_file(
                 number_of_voices -= 1
 
         result: list[str] = []
-        if len(pages) != len(svg_files):
-            total_svg_measures = sum(svg.number_of_measures for svg in svg_files) / number_of_voices
-            total_xml_measures = sum(page.number_of_measures for page in pages)
-            if total_xml_measures == total_svg_measures:
-                # This happens if the layout required extra pages
-                pages = [
-                    MusicXmlPage([], svg.number_of_measures // number_of_voices)
-                    for svg in svg_files
-                ]
-            else:
-                eprint(
-                    file,
-                    "INFO: Number of pages in SVG files",
-                    len(svg_files),
-                    "does not match number of pages in XML",
-                    len(pages),
-                )
+        assert len(pages) == len(svg_files)  # noqa: S101
 
-                return []
         for i, page in enumerate(pages):
             svg_file = svg_files[i]
             number_of_measures_per_voice_svg = svg_file.number_of_measures / number_of_voices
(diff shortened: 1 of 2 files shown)


weixlu:
And this coming Friday through Sunday is a public holiday here in China. I'm going to spend some time with my family.
Feel free to leave any comment!

liebharc:
Thanks for the PR @weixlu . I will take a look. Enjoy your holiday! Family time is really important :).

liebharc:
Amazing, you just increased the number of examples in the Lieder dataset by 50%. Well done :).

weixlu:
Ah, sad story, 😭. I just found the repeat sign detection may not be enough.
- This patch deals with this. (`datasets/Lieder-main/flat/lc4985931-1.svg`). The repeat sign here should be dropped.
image

- But I just found there is also another pattern. (`./datasets/Lieder-main/flat/lc5001870-1.svg`). The barline(also repeat sign) here should not be discarded. 
image

The solution should be, only discard barline if there is no note in between. So maybe I should first revert this patch and then start a new PR, is that OK? @liebharc

liebharc:
As you like. We can also just keep it and do another change on top of it. I don't think that there are any training runs scheduled right now. Germany is in a heat wave so the last thing you want to do is to run a GPU for 2-3 days :D.

weixlu:
OK, then let's just keep it. I also hear about the heat wave across euro. Back when BTC prices were high, I used to mine crypto on the PC in my bedroom. I still remember the heat and the noise — I could barely sleep!

PR #106, opened 2026-06-25
make empty staff visible in SVG

weixlu:
## Describe the bug
when running `training/datasets/convert_lieder.py`, this message (`Can't merge staffs with a different number of measures`) shows up 108 times, and this PR is intend to fix this. 
## Root Cause
This info appears when 2 staffs are paired incorrectly, thus can't be merged. The pair and merge procedure is done in `SvgMusicFile.merge_voice_with_next_one()`, it deals with the staff one by one, and relies on the assumption that all staffs are visible.
In fact, when rendering SVG, MuseScore can hide empty staff. For example, `datasets/Lieder-main/flat/lc5054946-4.svg`
image
## Expected result
To make `merge_voice_with_next_one()` works as expected, we need to make empty staff visible. After this PR is implemented, this image look like(Let's skip the first system, and only show the second one, e.g. measure14):

lc5054946-4-3.png:
image

lc5054946-4-3.tokens
```
clef_G2 _ _ _ _ upper
keySignature_2 . . . . .
rest_2. _ _ _ _ upper
barline . . . . .
```

lc5054946-4-4.png:
image

## How to fix
Just like what we have done in #94 , we need to change the 2 setting:
- the global setting: `<hideEmptyStaves>1</hideEmptyStaves>`
- the per-staff setting: `<hideWhenEmpty>1</hideWhenEmpty>`

Especially, there is a `cutaway` setting, which just hides part of the staff, e.g. some empty measures. Take `datasets/Lieder-main/flat/lc6623221-4.svg` for example
image
Solution for this case is similar, we simply discard the `cutaway` setting.

Unfortunatelly, the methods above do not cover every case. lc6236149, lc6209608 still fails, so we just skip them. And `datasets/Lieder-main/flat/lc6162644-2.svg` looks quite weird, I suggest skip it too. 
image

After all these fix, finally we only see this message `Can't merge staffs...` once in `lc6248301.musicxml`. I haven't seen it deeper, this sheet looks no-problem, not sure where is the root cause.

### DIFF
--- a/training/datasets/convert_lieder.py
+++ b/training/datasets/convert_lieder.py
@@ -124,7 +124,9 @@ def create_formats(source_file: str, formats: list[str]) -> list[dict[str, str]]
 
     # sq8940236: MuseScore seems to hang up
     # lc5001945: nested tuplet, not good for training
-    files_with_known_issues = ["sq8940236", "lc5001945"]
+    # lc6209608, lc6236149: empty&invisible staff in the very first system
+    # lc6162644: irregular staff in the last svg page
+    files_with_known_issues = ["sq8940236", "lc5001945", "lc6162644", "lc6209608", "lc6236149"]
     if any(issue in source_file for issue in files_with_known_issues):
         return jobs
     for target_format in formats:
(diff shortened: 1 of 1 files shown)


liebharc:
It's really amazing how you really looked into the details of the Lieder dataset @weixlu . Thanks!

liebharc:
FYI, I just made short 6 epoch run overnight to see how we are doing after all the changes to the dataset conversion. And it looks good!

```## Run 407 - at epoch 6

Commit: fabab4b0f8480be20d13686edf5f91d913ab00fd
Day: 26 June 2026
Transformer Smoke Test: 17%
System Level: 9.3 diffs, SER: 7.4%
Polish scores with musicdiff: OMR-NED 46.4%

Quick training run after updates to the dataset conversions. Results look okay for only 6 epochs.
```

weixlu:
> FYI, I just made short 6 epoch run overnight to see how we are doing after all the changes to the dataset conversion. 

Oh, you remind me of this. In fact I plan to further solve some problems in dataset, and then start training to see if things get better

weixlu:
And I found clean the dataset really requires some careful work, and may take longer than I expected...

liebharc:
Yes, it's actually hard work. And it's critical for the performance. Like any bad example will degrade performance. So if in doubt it's better to exclude examples from the training than to risk incorrect data.

I really appreciate your help here!

liebharc:
> > FYI, I just made short 6 epoch run overnight to see how we are doing after all the changes to the dataset conversion.
> 
> Oh, you remind me of this. In fact I plan to further solve some problems in dataset, and then start training to see if things get better

Quick training runs from time to time can be helpful. Otherwise if the training goes wrong then it will be hard to find the cause.

PR #112, opened 2026-07-01
update conversion script path

weixlu:
This is a trivial patch, just update conversion script path to `training/omr_datasets/convert_xxxx` in `training.md`


weixlu:
a little strange, I only edit `training.md`, but the ci failed on github (although it's ok on my side).

`Claude Code` suggest that this ci config may be wrong
```
cache-dependency-path: setup.py
```
because there isn't `setup.py` but `poetry.lock`

I follow his suggestion and then ci pipeline works.

liebharc:
Oh, that was my mistake to not update this properly. Thanks for fixing that!

PR #114, opened 2026-07-03
Use standardized (PEP 517) dependency syntax in pyproject.toml

geofft:
This is now the preferred format in Poetry's documentation https://python-poetry.org/docs/dependency-specification/ and it also allows using other tools such as uv for development.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,63 +1,72 @@
-[tool.poetry]
+[project]
 name = "homr"
-version = "0.1.0"
 description = "End-to-end Optical Music Recognition (OMR) system build on top of vision transformers."
-authors = ["Christian Liebhardt <christian.liebhardt@arcor.de>"]
+authors = [{name = "Christian Liebhardt", email = "christian.liebhardt@arcor.de"}]
 license = "AGPL-3.0"
 readme = "README.md"
 homepage = "https://github.com/liebharc/homr"
 documentation = "https://github.com/liebharc/homr"
+requires-python = "<3.16,>=3.11"
+dependencies = [
+  "numpy >= 2.4.2, < 3",
+  "opencv-python-headless >= 4.13.0.92, < 5",
+  "Pillow >= 12.1.1, < 13",
+  "types-Pillow >= 10.2.0.20240822, < 11",
+  "typing_extensions >= 4.15.0, < 5",
+  "musicxml == 1.4",
+  "rapidocr >= 3.7.0, < 4",
+  "onnxruntime >= 1.24.1, < 2",
+  "requests >= 2.32.5, < 3",
+]
+dynamic = ["version"]
 
-[tool.poetry.dependencies]
-python = "<3.16,>=3.11"
-numpy = "^2.4.2"
-opencv-python-headless = "^4.13.0.92"
-Pillow = "^12.1.1"
-types-Pillow = "^10.2.0.20240822"
-typing_extensions = "^4.15.0"
-musicxml = "1.4"
-rapidocr = "^3.7.0"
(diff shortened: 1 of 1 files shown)


geofft:
This allows `uv run homr` to work inside a git clone, and once it's merged `uvx git+https://github.com/liebharc/homr` will also work to allow building the latest commit from source. It also preserves the ability to work with Poetry, and in particular it continues to use Poetry as the build backend, even when you're using uv as the frontend.

I've double-checked that the changes in the `poetry.lock` file are very minimal, mostly just churn in the order of unordered lists. (At the moment I'm not checking in the `uv.lock` file, and I think so long as you're continuing to use Poetry as your primary development tool, you should continue to check in `poetry.lock` and not have anyone check in a `uv.lock` file.)

One small semantic change here is that this preserves `homr[gpu]` as an extra, i.e., a thing that gets encoded into the wheel, but moves `homr[dev]` and `homr[docs]` to dependency groups, i.e., these dependencies are not encoded in the wheel and you can only invoke them from local development. I _think_ this is the intended use case of these three groups, and doing it this way reduces the amount of churn in poetry.lock, but I'm happy to change this if you don't think it's right. Other than that, no semantic changes are intended.

If you try this out with uv, note there is a bug with the conflict around `opencv-python` vs. `opencv-python-headless` that means that if you're on a headless system uv may install the dependencies in the wrong order. I'll open a PR for that on top of this one in a second - uv has a way to suppress the unwanted `opencv-python` dependency. (If you're on a desktop system it should work fine, including with GPU support.)

cc @Alex23rodriguez, you might be interested since you opened #80.

geofft:
Actually I guess `homr[gpu]` currently _isn't_ an extra, which is why `uvx homr[gpu]`/`pip install homr[gpu]`/etc. don't work. So this has the side effect of making that work! I'll send another PR updating the documentation once I figure out whether we can do something about that `opencv-python-headless` issue for the non-development case.

weixlu:
@geofft Hi, good work! Thanks for your contribution.

Yes, typically I'm using `uv` in my daily work, so I'm quite happy to see it supported. It seems @liebharc may be more familiar with `poetry`, so I'd like to wait for his opinion on this.

weixlu:
Ah! I think we've met before! https://github.com/astral-sh/python-build-standalone/issues/1097

liebharc:
> It seems @liebharc may be more familiar with poetry, so I'd like to wait for his opinion on this.

It sounds like this PR enables us to use both Poetry and uv? I think that it's a plain improvement, and thanks for the contribution @geofft . In general, I would also be fine to move to uv as primary/only tool if that's a direction which you would recommend and if all the contributors agree (which I believe could be the case). homr is older than uv (not sure if the first public commit shows that) so back then it was simply not an option.

So fine for me @weixlu , I would leave the decision to hit the merge button to you.

weixlu:
I have checked the following things:
1. With this PR, `poetry` and `uv` are both available
2. changes in `poetry.lock` are just about orders
3. Yes, `homr[gpu]` should be included in the wheel, not `[dev]` nor `[docs]`. And currently `pip install homr[gpu]` doesn't work as expected. After your patch, it's now working!
4. About the conflict of `opencv-python-headless` and `opencv-python`, I'm using a headless linux, but I haven't come across the issue.

So all in all, I think this patch works perfect! Thanks again.

weixlu:
Ah, the CI failed. It seems we still need to regenerate `poetry.lock`, although it doesn't change anything, just the hash code or something.

Don't worry, I'll fix this #116

weixlu:
CI looks good for #116

Issue #110, opened 2026-06-30
Bug: Transformer generates an extra B3 note at the end of measure 3


JoeZhang42:
# Bug: Transformer generates an extra B3 note at the end of the 3rd measure of this page that doesn't exist in the source image

## Environment

- **homr version**: 0.7.0
- **Command**: `homr --coreml-encoder test_score02.jpg`
- **Hardware**: Apple M3, 16GB
- **Output format**: MusicXML

## Description

When processing a standard B♭ major piano score (digital score screenshot), the MusicXML output contains an extra B3 eighth note in the right hand (staff 1, voice 1) at the end of the 3rd measure of this page. This note does not exist in the original image.

## Location

- **Score**: test\_score02 (B♭ major piano score)
- **Measure**: 3
- **Part**: Right hand (staff 1, voice 1)
- **Position**: End of measure
- **Wrong note**: B3 (step=B, octave=3), eighth note, duration=2

## Context (correct musical passage)

The correct right hand sequence in measure 3 should be:

C5 (16th) → G4 (16th) → F4 (16th) → B♭4 (quarter) → ***(no note here)*** → measure 4 begins

But homr outputs an extra B3 eighth note after the B♭4 quarter, then proceeds to measure 4.

## Reproducibility

**100% reproducible.** Running homr twice on the same image produces the exact same wrong note at the exact same position. This indicates it is a systematic misrecognition, not a random hallucination.

## Root Cause Hypothesis

There are no musical notes at that position in the original image. However, other printed elements (fingering markers, pedal markings, dynamic signs, or layout artifacts) may have been misrecognized by the Transformer as a B3 eighth note.

## Expected Behavior

The model should not generate notes at positions where no notes exist in the source image.

## Attachment

Source image file: `test_score02.jpg`

(The image is a screenshot of an digital piano score in B♭ major, containing melody, chord accompaniment, pedal markings, etc.)

<img width="1078" height="1424" alt="Image" src="https://github.com/user-attachments/assets/87894156-0a8c-43c2-a846-b6ce7d6dd5be" />

liebharc:
Hi thanks for the report! To set overall expectations: OMR systems aren't perfect, state-of-the-art error rates reported in literature are usually still >5% for real world scenarios, which means that 1 out of 20 symbols has an error. That's also what you can reasonably expect from homr.

[...]

PR #115, opened 2026-07-05
MuseScore allows to change the y-position of a note

liebharc:
E.g. in lc4926375.mscx that causes a E5 note to be positioned one step down so that it looks like a D5 note. As this degrades performance, we remove that from the training data.


liebharc:
Since this seems to improve results on #110 , I would go ahead and merge it. Keep in mind that the lieder dataset needs to be regenerated for this to have an effect.

PR #119, opened 2026-07-13
avoid decode/encode overhead in benchmark dataset loading

weixlu:
## Motivation

When I run:

```bash
poetry run python -m validation.smb --tool homr
```

it takes about two minutes after pressing Enter before the following message appears:

```text
--- Batch 1/69 (1–10 of 685)
```

In other words, there is a roughly two-minute delay before the benchmark actually starts.

This PR optimizes this startup process and reduces 2 min to 6s (for `smb`). For `polish-scores` dataset, that's 45s to 6s.

## How It Works

The current dataset loading first uses `load_dataset` to decode the binary image data into Python image objects. It then calls `.save(PATH)` to encode the image back into binary form and write it to disk. Clearly, this is unnecessary.

Instead, we can just retrieve the raw bytes directly from the dataset and write them to disk. Then no decode-and-encode anymore.

## Other Changes

As I remember it, the benchmark-related code seems to be generated with `Claude Code`—please correct me if I am mistaken. I also use `Claude` or `ChatGPT` myself, and I have noticed that its code is often more verbose than necessary, even if it's functionally correct. I generally prefer simpler code that is easier to read and understand.

Therefore, I also simplified and refactored the dataset-related logic in benchmark. There are two main changes that are worth reviewing:

### 1. Dataset Split Selection

I noticed that SMB only has a `test` split, while `polish-scores` only has a `train` split. Since there is no actual choice of split for either dataset, I removed the `--split` option.

### 2. Dataset Iteration

The current dataset-loading code uses `yield` to return an `iterator`. Iterator(Generators) are quite advanced skills in python, useful in more complex producer-consumer or streaming scenarios.

But here, the dataset can simply be loaded and returned directly, which makes the control flow easier to follow. So from my understanding, I'd remove the iterator-based design. If I have overlooked something—please let me know.

### DIFF
--- a/validation/ned_benchmark.py
+++ b/validation/ned_benchmark.py
@@ -41,7 +41,7 @@ import signal
 import statistics
 import sys
 import traceback
-from collections.abc import Callable, Iterable, Iterator
+from collections.abc import Callable, Iterator
 from pathlib import Path
 
 from homr.transformer.vocabulary import EncodedSymbol
@@ -240,7 +240,7 @@ def update_ned_scores(
 
 
 def run_benchmark(
-    samples: Iterable[tuple[str, str, Path | None]],
+    samples: list[tuple[str, str, Path]],
     tool: Callable[[str, Path | None], str],
     limit: int | None = None,
     verbose: bool = False,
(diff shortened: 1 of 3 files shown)


liebharc:
@weixlu , thanks! I did a couple of test runs and it looks fine.

PR #120, opened 2026-07-14
update doc

weixlu:
- README.md: after #114 , gpu dependency should be installed via `--extras gpu`
- Training.md: update SMB benchmark result on Run 426. [smb_new.log](https://github.com/user-attachments/files/29996040/smb_new.log)
- Benchmark.md: That's what I want to discuss. Should we update it as well? And besides, I notice benchmark results are uploaded to [release page](https://github.com/liebharc/homr/releases/tag/benchmark), so maybe I should do the same this time. If so, then we need to think about how to give it a name for each run.

### DIFF
--- a/Benchmark.md
+++ b/Benchmark.md
@@ -131,7 +131,7 @@ The display uses `<` for deleted tokens (in reference, missing from output), `>`
 
 ## Benchmark results
 
-All results are mean OMR-NED in % over the samples where the tool produced output. Means are only comparable between tools when coverage is similar.
+All results are mean OMR-NED in % over the samples where the tool produced output. Means are only comparable between tools when coverage is similar. For performance on latest homr model, refer to `Training.md` for details.
 
 ## native parser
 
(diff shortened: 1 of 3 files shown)


liebharc:
Regarding, Benchmark.md and our release page. That's a good point. I think I want to keep things lean and not add too much process. In that sense I would occasionally update the numbers in Benchmark.md and upload new dbs to the release page, but not with every model update. It might be something to consider with every new minor version release. Training.md then remains the main spot to track results and this way we should also be able to spot regressions. And Benchmark.md and the DB releases have the purpose of giving people an idea what performance they can expect from homr.

Then it might however be good to add a note to Benchmark.md that these aren't necessarily the newest results.

weixlu:
Sure, I also think we don't have to add too much process. Then this PR may seem fine.

liebharc:
Yes, feel free to hit the merge button!

PR #122, opened 2026-07-15
Fixed staff detection errors

liebharc:
Fixes some staff detection problems which came up in the smb benchmark.

Disclaimer: Claude code changes which mostly accumulated while I was running it in the background, still have to review this myself

Performance changes

```
# Start

System level: 5.5 diffs, SER: 4.1%
Polish scores 24.30%
SMB 23.4%

# Staff parsing

System level: 5.6 diffs, SER: 4.7%
Polish scores 21.40%, 42.63%
SMB 19.98%

# Fix NED scoring

Polish scores 19.98%
SMB 16.60%

# Realized that polish ground truth is missing articulations

Polish scores 17.24% 40.23%

# Undone some claude code changes which looked like poorly justified

Polish Scores: 18.1% 42.10%
SMB: 14.5%
```

It turns out that SMB is also unreliable. For some pieces it omits staccato information which penalizes any model which correctly detect them. I decided to just keep it like that as it's likely not meaningful to spend the time on that.

### DIFF
--- a/homr/staff_detection.py
+++ b/homr/staff_detection.py
@@ -453,12 +453,6 @@ def find_staff_anchors(
                     if (line.max_x - line.min_x)
                     > constants.is_short_connected_line(estimated_unit_size)
                 ]
-            if len(connected_lines) == constants.number_of_lines_on_a_staff - 1:
-                extrapolated = _extrapolate_missing_staff_line(
-                    connected_lines, symbol, estimated_unit_size
-                )
-                if extrapolated is not None:
-                    connected_lines = extrapolated
             if not len(connected_lines) == constants.number_of_lines_on_a_staff:
                 continue
             if not are_lines_parallel(connected_lines, estimated_unit_size):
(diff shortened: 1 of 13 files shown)


weixlu:
Wow, this improvement looks significant!

homr/staff_detection.py
```diff
@@ -373,6 +453,12 @@ def find_staff_anchors(
                     if (line.max_x - line.min_x)
                     > constants.is_short_connected_line(estimated_unit_size)
                 ]
+            if len(connected_lines) == constants.number_of_lines_on_a_staff - 1:
```
liebharc: that sounds risky: The whole "anchor" detection is conservative as the anchors are really used as robust and reliable points from which the rest of the staff can be constructed. If you allow to guess anchors then you might end up detecting e.g. ledger lines as staff lines or slurs. And then the later dewarp step will actually distort the image.

homr/staff_detection.py
```diff
@@ -713,6 +864,7 @@ def detect_staff(
     )
 
     staff_anchors = filter_unusual_anchors(staff_anchors)
+    cap_anchor_zones_to_neighbors(staff_anchors)
```
liebharc: Would be interesting to see how much of a difference cap_anchor_zones_to_neighbors actually does.

PR #124, opened 2026-07-17
Allow invisbile tuplets again

liebharc:
Tuplets without markings are frequent in orchestral music. I would like to see if after all the cleanups we did we might be able to see if the model deals better with them.

### DIFF
--- a/training/omr_datasets/convert_lieder.py
+++ b/training/omr_datasets/convert_lieder.py
@@ -215,55 +215,6 @@ def _music_font_for_file(source_file: str) -> str:
     return _MUSIC_FONTS[index]
 
 
-"""
-In mscx file, tuplet can be set invisible via the following ways:
-1. <Tuplet( id="...")>
-     <visible>0</visible>
-   </Tuplet>
-    this is equal to <notations print-object="no"> in musicXML
-2. <numberType>2</numberType>
-    this is equal to <tuplet show-number="none" ...> in musicXML
-3. <bracketType>2</bracketType>
-    this is equal to <tuplet bracket="no" ...> in musicXML
-4. numberType & bracketType can be set globally via <tupletNumberType> & <tupletBracketType>
-
-recommend reading lc5033057 from Lieder to better understand the rules about tuplet visibility.
-"""
-
-
-def _make_tuplet_visible(mscx_file: str) -> None:
-    with open(mscx_file, encoding="utf-8") as f:
-        content = f.read()
-
-    # match: <visible>0</visible> or <numberType>2</numberType>
-    # or <bracketType>2</bracketType>
-    strip_re = re.compile(
-        r"[ \t]*<(?:visible>0|numberType>2|bracketType>2)"
-        r"</(?:visible|numberType|bracketType)>[ \t]*\r?\n"
-    )
-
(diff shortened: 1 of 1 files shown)


weixlu:
Hi, is the training ongoing? If not yet, I can rebuild the dataset and train

liebharc:
Hi @weixlu , not I haven't started training for this yet. Sure, would be curious how this PR does - although I fear that it will just start hallucinating triplets like crazy. But perhaps it works. So  thanks for training this one.

weixlu:
I got an error during training. After some investigation, I find out that `datasets/Lieder-main/flat/sq15049456-32-17.tokens` has a strange `note_1 C4 _ accent_fermata_trill slurStart upper`, where `accent_fermata_trill` can't be recognized. For a quick fix, I just remove `sq15049456-32-17.tokens` in `index.txt`

error message:
```
[rank0]: Original Traceback (most recent call last):
[rank0]:   File "/root/homr/.venv/lib/python3.12/site-packages/torch/utils/data/_utils/worker.py", line 349, in _worker_loop
[rank0]:     data = fetcher.fetch(index)  # type: ignore[possibly-undefined]
[rank0]:            ^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/root/homr/.venv/lib/python3.12/site-packages/torch/utils/data/_utils/fetch.py", line 52, in fetch
[rank0]:     data = [self.dataset[idx] for idx in possibly_batched_index]
[rank0]:             ~~~~~~~~~~~~^^^^^
[rank0]:   File "/root/homr/training/transformer/data_loader.py", line 88, in __getitem__
[rank0]:     tokens = self._read_tokens(tokens_full_filepath)
[rank0]:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/root/homr/training/transformer/data_loader.py", line 58, in _read_tokens
[rank0]:     return to_decoder_branches(tokens)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/root/homr/training/transformer/training_vocabulary.py", line 291, in to_decoder_branches
[rank0]:     articulations.append(vocab.articulation[symbol.articulation])
[rank0]:                          ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
[rank0]: KeyError: 'accent_fermata_trill'
```

liebharc:
The error comes from the vocabulary. The articulation branch is directly generated from the training data. So everytime there is a change to the dataset it's worth to run https://github.com/liebharc/homr/blob/main/training/transformer/training_vocabulary.py as a script to check if every token is still supported (and also if it's still needed). Any missing tokens should then be added to the vocab. Removing the file is also completely fine for this round.

weixlu:
Oh, thanks for pointing out. So vocabulary should also be updated along with dataset

liebharc:
@weixlu , thanks for doing the training run! The model behaves exactly as before this change. So no improvement in tuplet detection but also no hallucinations. I think I would go forward and still merge this, as it means that we can remove a bit of code.

weixlu:
Yes, of course. Less code makes the control flow simpler

PR #126, opened 2026-07-20
Fix missing rest in XML

weixlu:
## Introduce the issue
I would like to discuss a limitation of the current token format. The `.tokens` in homr distinguishes between the treble (marked as `upper`) and bass (marked as `lower`) staves in a piano score, but it does not distinguish between multiple voices within the same staff. For example, the two images below produce identical `.tokens` files.
image
image
Is this behavior intentional, or is it something that could be improved?

## My thought
Personally, I think the current behavior is acceptable, since the case shown above is quite rare. Properly distinguishing multiple voices within the same staff would require big changes to the token format. homr already handles more common cases involving multiple voices on the same staff. see the first measure in the example below
- original input
ori

- generated by homr
gen-base

## What this PR did
Therefore, instead of making such a large change, this PR proposes a small fix specifically for handling rests.

See the last measure in the example above, the rest is missing. I found that transformer actually detected the rest and included it in the token sequence. However, it is dropped while generating the XML. So I found the bug and fix it, now the rest is preserved as shown below.
gen-fix

Another benefit of this fix is, `convert_musetrainer` previously print `SIGSEGV` because the generated XML was invalid. Now that the XML generation issue has been fixed, the segmentation fault no longer occurs.

### DIFF
--- a/homr/music_xml_generator.py
+++ b/homr/music_xml_generator.py
@@ -1,3 +1,5 @@
+# flake8: noqa: S101
+
 import math
 import xml.etree.ElementTree as ET
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


liebharc:
That's a real issue I believe. Here example 37 from polish scores:
grafik

The first E appears twice, once as a quarter note and once as an eight note. homr only detects the quarter note. The limitation really is in the vocabulary as you found. Your fix looks like an improvement (I will need to take a 2nd look later). Not sure what it would require to get the generic case handled correctly. humdrum supports this with multiple parallel spines and musicxml has backups to encode multiple voices in the same staff. I think both aren't really suitable for a transformer as there is ambiguity on how you encode symbols (e.g. which voice comes first?). Possibly we would need to detect stem direction and flags to reconstruct the voices correctly. But that is likely beyond this PR. I will try to take another look soon!

weixlu:
Ah yes, I have seen something similar to the example you provide. So this is real case. I also haven't figured out a good way to handle this. And it seems not easy.

liebharc:
The PR seems to work well. I reran System Level and Polish Scores and looked through the changes. All good. Feel free to hit merge when you are ready.

PR #127, opened 2026-07-22
increase timeout for multi-gpu communication

weixlu:
Now dataset convertion is added to training pipeline, so we need to increase timeout(default 10min, increased to 3 hours) for multi-gpu communication. Otherwise GPU0 may be doing dataset convertion for a long time, while GPU1 is waiting such a long time that it just timeout
```python
    if distribute.is_rank0():
        _check_datasets_are_present(dataset_index) # take a long time
    distribute.barrier() # timeout
```

### DIFF
--- a/training/transformer/distribute.py
+++ b/training/transformer/distribute.py
@@ -2,6 +2,7 @@
 # flake8: noqa: S101
 
 import os
+from datetime import timedelta
 
 import torch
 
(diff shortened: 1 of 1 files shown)


liebharc:
Hi @weixlu , you are likely the only one with a multi GPU setup so I can't say little about the changes. Feel free to hit merge when you are ready.

PR #128, opened 2026-07-22
limit OpenCV thread count

weixlu:
During transformer training, we have one main process and 12 dataloader processes. After #107 , data augmentation is also performed inside the dataloader, which uses the OpenCV lib (`cv2`).

However, OpenCV itself is still multithreaded. Assuming our machine has 16 CPU cores, OpenCV may create 16 threads inside each dataloader process. This is problematic because the dataloader is already parallelized (12 processes). If each dataloader performs additional parallel work (for example, using 16 OpenCV threads), the total number of threads becomes roughly `12 × 16 = 192`, which is far beyond what our CPU can efficiently handle.

Therefore, I suggest disable OpenCV's internal multithreading.

```python
cv2.setNumThreads(0)
```

I tested this on my machine with both single-GPU and multi-GPU training. The results show that disabling OpenCV multithreading does not affect training speed, while CPU utilization is significantly reduced.

### DIFF
--- a/training/transformer/image_utils.py
+++ b/training/transformer/image_utils.py
@@ -12,6 +12,7 @@ from homr.type_definitions import NDArray
 from training.transformer.augraphy_augment import apply_augraphy
 
 _texture_config: dict = {"dir": None, "paths": None}
+cv2.setNumThreads(0)
 
 
 def set_texture_dir(path: str) -> None:


weixlu:
I also want to ask how `dataloader_num_workers=12` was determined. It looks like this was introduced in an early patch. My understanding is that this number is highly dependent on the hardware. If it has 16 cores (or 8 cores/16 threads) or more, then 12 workers seems reasonable. However, if the CPU only has 6–8 cores (for example, my gaming PC), then this value is obviously too large. Maybe slightly descrease it won't slow down the training speed? (just a guess)

liebharc:
Hi @weixlu , I haven't tested this (I'm on a different machine), The explanation makes a lot of sense so feel free to merge it.

> I also want to ask how dataloader_num_workers=12 was determined. 

You guessed everything correctly - including the number of cores on my CPU. This parameter, the batch size of the training and perhaps some other params which I forgot are all set to values which work for my machine. For a long while I was the only one who trained the transformer so that was a reasonable start. 

We could also expose the values as args for the training. Or define a machine local config file (and in .gitignore) which allows to overwrite the values.

weixlu:
Let me merge it since it's a trivial patch. If you see any regression later, please tell me and I'll fix it.

About these training params, I plan to detect the hardware config then set them dynamically. Maybe a next PR will do this.

PR #130, opened 2026-07-28
Benchmark optimization

aicelen:
Optimizes two things:
1. Stop reloading the segnet every homr call
2. Make title detection optional

The first 10 examples of polish score now need 65s instead of 71s (tested without multi-threading but this shouldn't have any impact on the inference time)

Edit: Could one of you take a quick look if I missed something?


weixlu:
Hi, your optimization ideas look good! About title detection, I think the code can be simplified: change `title_future = detect_title(debug, staffs[0])` to 
```python
if config.title_detection:
  title_future = detect_title(debug, staffs[0])
else:
  title_future: Future[str] = Future()
  title_future.set_result("")
```
Then we don't need to submit an empty job for executor, i.e., don't need to change `homr/title_detection.py`

weixlu:
Also, it seems you have to use 
`["poetry", "run", "python3", "-m", "homr.main", str(image_dir), "--no-title"]`
instaed of 
`["poetry", "run", "python", str(_HOMR_MAIN), str(image_dir), "--no-title"]`

That's a little strange. What error message did you see? Looks like the diference is just `python3` vs `python`. `str(_HOMR_MAIN)` is just the path of `main.py`, so it should work.

aicelen:
>About title detection, I think the code can be simplified

Yes that is way cleaner

>What error message did you see?
```
[0] ERROR: homr exited with code 1
Traceback (most recent call last):
  File "/home/enno/Documents/GitHub/homr/homr/main.py", line 14, in <module>
    from homr import color_adjust, download_utils
ModuleNotFoundError: No module named 'homr'
```
I use `python3 -m validation.polish-scores --tool homr --limit 10 --workers 4`

Just run the whole polish-scores benchmark which took 220s, so this seems quite fast.

Why do we use a batch_size? I tested with batch_size=5 and batch_size=20 which both use the same amount of memory. If a batch is finished we reload the homr model which costs time. Couldn't we introduce an automatic value that is test_samples/workers?

weixlu:
According to the error message, my guess is, maybe you can run `poetry install` to fix this. But anyway, `poetry run python3 -m homr.main -h` also works, (at least on my side), so I think we can just use it to replace `["poetry", "run", "python", str(_HOMR_MAIN), str(image_dir), "--no-title"]`.

weixlu:
About batch_size, I have read related code. I think it's OK either use it or not.
- Yes, it won't reduce memory usage, because the images are handled one by one. Base on this, it looks useless.
- But it won't spend much time reloading model, my guess is about 1 or 2 seconds. Base on this, it looks harmless.

aicelen:
Alright thank you for the review. I agree with you that reloading the model is harmless, so we'll leave the arg based batch size for now.

PR #132, opened 2026-07-29
Fix and clean-up the colab notebook (#131)

### LINKED ISSUE #131: colab.py line not working?
**Describe the bug**
I run the colab file today in colab, and got this
for file_name in filenames:
for file_name in filenames:
    !poetry run homr "{file_name}"
Skipping virtualenv creation, as specified in config file.
Command not found: homr

Image

I have ran this before and this is first time I got a problem.

aicelen:
This fixes #131 
- the poetry install command wasn't updated after #114
- colab misses the poetry-dynamic-versioning plugin; I don't know why but the log said:

```
The following Poetry plugins are required by the project but are not installed in Poetry's environment:
  - poetry-dynamic-versioning[plugin]
Installing Poetry plugins only for the current project...
Path /colabtools/dist/google_colab-1.0.0.tar.gz for google-colab does not exist
```

I also removed previous outputs so it is clearer and easier to use.


aicelen:
Alright this is already fixed.

PR #135, opened 2026-08-03
support PDF as input format

weixlu:
The current input format only supports image file(PNG/JPG). PDF is also a common format, especially for multi-page scores, which can conveniently be bundled into a single PDF file.

This PR adds support for PDF. It's easy to do this: let PDF rendered into PNG, then processed as usual.

I used `pypdfium2 5.12.1` to read PDF, as this lib appears to have a permissive license (Apache-2.0). Anyway, if we encounter any licensing concerns later, it should be easy to replace it with another one.

I tested it with two PDF scores generated by MuseScore, and the conversion appears to work correctly.


liebharc:
The changes look good. I didn't add PDF support so far, because with some PDFs you can also be smarter: If the PDF contains a vector graphics then it gets much easier to e.g. extract pitch information by parsing the vector information and not use a raster format such as PNG. I believe Audiveris can do this.

Feel free to merge when you are ready! Also, I will start a new job in Sept (which is pretty exciting!), so I might need more time to respond in the future.

weixlu:
Congrats on the new job! I wonder what else might be out there too, especially recently we are asked to stay an extra hour every day. (Hopefully my boss never sees this.😂 Maybe we really should follow enno's suggestion and create a private Discord channel )

PR #137, opened 2026-08-07
Add option to export pixel coordinates for noteheads

lennart-rth:
Add the flag `write_noteheads_json` to `ProcessingConfig`. 
When set to true, we export the coordinates and additional information into a JSON file. 
This can be useful when any new information in the original file needs to be augmented or added based on the OMR detection.

### DIFF
--- a/homr/main.py
+++ b/homr/main.py
@@ -273,6 +273,20 @@ def process_image(
             with open(json_file, "w", encoding="utf-8") as f:
                 json.dump(output_data, f, indent=2)
             eprint("Notehead coordinates written to", json_file)
+
+            json_staffs_file = replace_extension(image_path, "_staffs.json")
+            staffs_data = []
+            for staff_idx, multi_staff in enumerate(multi_staffs):
+                for staff in multi_staff.staffs:
+                    staffs_data.append({
+                        "staff_index": staff_idx,
+                        "top_y": float(staff.min_y),
+                        "bottom_y": float(staff.max_y),
+                        "spacing": float((staff.max_y - staff.min_y) / 4.0)
+                    })
+            with open(json_staffs_file, "w", encoding="utf-8") as f:
+                json.dump({"staffs": staffs_data}, f, indent=2)
+            eprint("Staff positions written to", json_staffs_file)
         debug.write_teaser(teaser_file, multi_staffs)
         debug.clean_debug_files_from_previous_runs()
 
--- a/homr/main.py
+++ b/homr/main.py
@@ -1,8 +1,8 @@
 import argparse
 import glob
+import json
 import os
 import sys
-import json
 import xml.etree.ElementTree as ET
 from concurrent.futures import Future
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


homr/main.py
```diff
@@ -228,6 +226,67 @@ def process_image(
...
             staff_position_files = replace_extension(image_path, ".txt")
             save_staff_positions(multi_staffs, image.shape, staff_position_files)
+        if config.write_noteheads_json:
+            json_file = replace_extension(image_path, "_coordinates.json")
+            noteheads_data = []
+            note_id = 1
+            for staff_idx, multi_staff in enumerate(multi_staffs):
+                for staff in multi_staff.staffs:
+                    for note in staff.get_notes():
+                        box = note.box
```
liebharc, line 236: You likely don't want to use this info. It's the bounding box from the segmentation prediction and basically unused. More likely you want to go through the detected symbols and use their coordinates.

For background take a look at the README: 

```Note: The transformer output provides the sequence of symbols but does not include explicit positional information (horizontal or vertical coordinates). However, the model computes the center of attention as a byproduct of the attention mechanism, which can be used to estimate the focus point on the staff image.```

And the class EncodedSymbol in vocabulary.py

**Comment by liebharc** [comment]
I left you one comment regarding your approach. In general, it would be also helpful to better understand your use case. Adding a new option is cheap, but it costs us in the long run as we have to maintain it. So it would be just good for us to understand why it's needed.

File: homr/main.py [hunk 2]
```diff
@@ -228,6 +226,67 @@ def process_image(
...
+                        noteheads_data.append(
+                            {
+                                "id": note_id,
```
liebharc, line 239: not sure why you need the ID, if it's just the index of the JSON array. Perhaps it could be removed?

homr/main.py
```diff
@@ -228,6 +226,67 @@ def process_image(
...
+                                "staff_index": staff_idx,
+                                "center": [float(box.center[0]), float(box.center[1])],
+                                "size": [float(box.size[0]), float(box.size[1])],
```
liebharc, line 242: Everything except center is specific to the bounding box attention. The center of attention will only give you the center itself.

homr/main.py
```diff
@@ -228,6 +226,67 @@ def process_image(
...
+            output_data = {
+                "image_path": image_path,
+                "total_noteheads": len(noteheads_data),
+                "noteheads": noteheads_data,
+            }
+            with open(json_file, "w", encoding="utf-8") as f:
+                json.dump(output_data, f, indent=2)
+            eprint("Notehead coordinates written to", json_file)
+
+            json_staffs_file = replace_extension(image_path, "_staffs.json")
```
liebharc, line 277: Here the new option and write_staff_positions seem to overlap. You likely have a specific use case min mind, but in general that will confuse users. Instead: Can we just write the notehead positions with the new option and nothing else? And then if you need additional staff info extend the other option?

homr/main.py
```diff
@@ -228,6 +226,67 @@ def process_image(
         if config.write_staff_positions:
             staff_position_files = replace_extension(image_path, ".txt")
             save_staff_positions(multi_staffs, image.shape, staff_position_files)
+        if config.write_noteheads_json:
+            json_file = replace_extension(image_path, "_coordinates.json")
```
liebharc, line 230: Please extract a method, analog to save_staff_positions to avoid having this long code block in process_image

liebharc (changes requested):
I left some more comments, I was in a hurry, so the tone might be more direct. Apologies for that.

Issue #139, opened 2026-08-21
Would you accept PRs for <accidental>, <tie> and <beam> in the MusicXML export?


pschuegr:
Asking before writing code, since two of these are clear-cut and one is your preference.

I've been using homr to build known-good MusicXML encodings of public-domain
piano scores. The recognition has been excellent — on a MuseScore engraving of
Satie's Gymnopédie No. 1, comparing homr's raw per-page output against the
source encoding it was rendered from:

|  | source | homr |
|---|---|---|
| `<note>` | 557 | **557** |
| `<alter>` | 100 | **100** |
| `<accidental>` | 43 | 0 |
| `<tie>` | 28 | 0 |
| `<tied>` | 28 | 0 |
| `<beam>` | 0 | 0 |

Every note and every alteration, exactly right. So this issue is only about
what the XML writer records, not about recognition.

Tested on `0.7.0.post34+2d0c0a6` (main); same on 0.7.0 from PyPI.

## 1. `<accidental>` is never emitted

`<alter>` gives the sounding pitch and homr sets it correctly. `<accidental>`
is the separate element saying a sign is *printed*, and `music_xml_generator.py`
never emits one.

This is invisible in some tools and glaring in others: MuseScore infers the
printed accidental on import, so files look fine there, while Verovio maps
`<alter>` to a gestural accidental only and renders a score with no sharps at
all — correct playback, unreadable page.

To be clear this isn't #49, which was the transformer dropping the `#` from a
repeated F♯ and so changing the sounding pitch. This is downstream of the
model, in the writer.

Which sign each note needs follows from the key signature and what the bar has
already altered, so no extra recognition is involved. I implemented it as a
post-processor over homr's output and checked it against the source encoding
above: 43 expected, 43 derived, all agreeing on measure, note and sign — none
missed, none spurious.

## 2. `<tie>` is never emitted, only `<tied>`

At `music_xml_generator.py:520-523`:

```python
elif articulation == "tieStart":
    ET.SubElement(notation, "tied", type="start")
elif articulation == "tieStop":
    ET.SubElement(notation, "tied", type="stop")
```

`<tied>` inside `<notations>` draws the curve; `<tie>` as a direct child of
`<note>` is the sounding tie. Only the first is written, so whenever a tie
*is* detected, the result shows the slur-like curve but still re-strikes the
note on playback.

This looks like a two-line addition next to the existing branches, and it's
independent of how well ties are detected — worth fixing either way, I'd have
thought.

[...]

PR #141, opened 2026-08-22
Recover ties from same-pitch slurs

pschuegr:
Refs #139. Implements the suggestion from @aicelen there: since the model
learns slurs and ties as one class, a tie can be read back out of the slurs.

A slur is treated as a tie when it joins a note to the very next note of its
own voice and both carry the same pitch. `convert_ties()` runs once per part
from `build_part()`, after its measures are built — per part rather than per
measure because ties cross barlines constantly.

**No pairing of slur starts with stops.** `slur_number` is the staff number, so
every slur on a staff shares a number and they cannot be told apart by it — I
found four open at once on one test file. Any pairing rule would be guessing,
and a tie needs none: it is a property of a note and its immediate successor.
`build_slurs()` writes at most one start and one stop per note, so which
element to remove is never ambiguous either.

A converted pair gets **both** `<tie>` and `<tied>`. Only `<tied>` was written
before, so a detected tie drew as a curve while both notes still sounded
separately — @weixlu reports running into that too.

## Results

Measured against known-good encodings. Engraved input, where curve detection
is good:

| source | ties | found | correct | recall | precision |
|---|---|---|---|---|---|
| Bach, WTC I Prelude in C (BWV 846) | 68 | 65 | 63 | 0.93 | 0.97 |
| Satie, Gymnopédie No. 1 | 14 | 14 | 14 | 1.00 | 1.00 |

Scanned editions, against Craig Sapp's kern, where it is not:

| source | ties | found | correct | recall | precision |
|---|---|---|---|---|---|
| Joplin, Heliotrope Bouquet | 189 | 31 | 23 | 0.12 | 0.74 |
| Chopin, Mazurka Op. 41 No. 2 | 85 | 3 | 3 | 0.04 | 1.00 |

A Beethoven scan with 22 slur pairs and no ties gains none.

**The scan figures are detection, not conversion.** On the Joplin page carrying
51 ties the model emits about 25 curves; on the Chopin it finds roughly 77
curves for 123 real ones, but their endpoints rarely land on two adjacent notes
of one pitch. Resolution is not the cause — 2538px and 5076px inputs give
identical output. This is the weakness described in #81, measured on scans.

The Chopin also has 38 phrase slurs and none of them was converted, which is
the failure mode I most wanted to rule out.

BWV 846's five misses are all in the closing bars, where the model puts a slur
start on a treble note and its stop two staves below — not a tie by any
reading. Relaxing the rule from "next note in this voice" to "next note on this
staff" recovers none of them, so it is already extracting what the model
provides.

So: ties are recovered wherever the curve is detected accurately, and nothing
is invented where it is not.

## Tests

Six cases added to `tests/test_music_xml_generator.py`: same pitch becomes a
tie; different pitch stays a slur; a tie across a barline; same pitch but not
adjacent stays a slur; a part with no slurs is untouched; and a tie nested
inside a longer slur on the same staff, which is the case the shared slur
number makes ambiguous.

`make ci` passes — 73 tests, ruff, black, isort and mypy clean.

The `<accidental>` and `<beam>` questions from #139 are deliberately left out
of this PR so they can be taken or refused separately.

### DIFF
--- a/homr/music_xml_generator.py
+++ b/homr/music_xml_generator.py
@@ -454,30 +454,67 @@ def convert_ties(part: ET.Element) -> None:
     property of a note and its immediate successor.
 
     Run per part rather than per measure, because ties cross barlines. One
-    pass is enough: the last note seen in a voice is by definition the
+    pass is enough: the last event seen in a voice is by definition the
     predecessor of the next one in it.
     """
-    previous: dict[tuple[str, str], ET.Element] = {}
+    previous: dict[tuple[str, str], list[ET.Element]] = {}
     for measure in part.findall("measure"):
-        for note in measure.findall("note"):
-            key = (note.findtext("staff", "1"), note.findtext("voice", "1"))
+        for event in group_into_events(measure):
+            first = event[0]
+            key = (first.findtext("staff", "1"), first.findtext("voice", "1"))
             before = previous.get(key)
-            previous[key] = note
+            previous[key] = event
             if before is None:
                 continue
-            pitch = get_note_pitch(before)
-            if pitch is None or pitch != get_note_pitch(note):
+            tie_event(before, event)
+
+
+def group_into_events(measure: ET.Element) -> list[list[ET.Element]]:
+    """The measure's notes, with each chord kept together as one event.
+
+    A chord is one moment of the music, not several: <chord> means "sounds
+    with the note before". Walking notes one at a time makes a chord's own
+    members each other's neighbours, so nothing in a chord ever finds the note
+    it is really adjacent to, and no tie between two chords is ever seen.
+    """
+    events: list[list[ET.Element]] = []
+    for note in measure.findall("note"):
+        if note.find("chord") is not None and events:
+            events[-1].append(note)
+        else:
+            events.append([note])
+    return events
+
+
+def tie_event(before: list[ET.Element], after: list[ET.Element]) -> None:
+    """Convert every slur between two adjacent events that is really a tie.
+
+    Matched by pitch, and led by the slurs: the model marks the notehead the
+    curve touches, which on a chord is one member and not the whole thing -
+    of the chords carrying a slur on my test material, 25 of 26 had it on
+    some members only. So each slurred note looks for its own pitch in the
+    next event rather than the chord being taken as a whole.
+    """
+    for start_note in before:
+        begins = get_slur(start_note, "start")
+        if begins is None:
+            continue
+        pitch = get_note_pitch(start_note)
+        if pitch is None:
+            continue
+        for stop_note in after:
+            if get_note_pitch(stop_note) != pitch:
                 continue
-            begins = get_slur(before, "start")
-            ends = get_slur(note, "stop")
-            if begins is None or ends is None:
+            ends = get_slur(stop_note, "stop")
+            if ends is None:
                 continue
             begin_slur, begin_notation = begins
             end_slur, end_notation = ends
             begin_notation.remove(begin_slur)
             end_notation.remove(end_slur)
-            add_tie(before, "start", begin_notation)
-            add_tie(note, "stop", end_notation)
+            add_tie(start_note, "start", begin_notation)
+            add_tie(stop_note, "stop", end_notation)
+            break
 
 
 def build_clef(model_clef: EncodedSymbol, attributes: ET.Element) -> None:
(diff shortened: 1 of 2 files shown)


weixlu:
Hi, thanks for your contribution! I tested some scores on your branch, and ties come out correctly. So the implementation works — I just think the code could be simplified. 
I'd perfer to place tie detection on `list[SymbolChord]` in `build_measures()`, like `add_tuplet_start_stop()`. At that point a `SymbolChord` already represents the time index, so we don't need to build `latest` or `successor` like below.
```python
    successor: dict[int, int] = {}
    latest: dict[tuple[str, str], int] = {}
    for index in reversed(range(len(notes))):
        note = notes[index]
        key = (note.findtext("staff", "1"), note.findtext("voice", "1"))
```

Also, I notice `convert_ties()` scans backwards first and then forwards. I'm not sure why we need two passes — one pass looks enough to me.

weixlu:
The new code looks good, and I verify it on some scores. Thanks! I'd suggest holding off on merge until enno or liebharc has a chance to take a look.

pschuegr:
Thanks! Will wait for further review.

FYI on the `SymbolChord` idea, since I did try it: the voice isn't available at
that point. `into_positions()` splits only on `position`, which is
`upper`/`lower` — the staff. The voice is made later, in `build_note_chord()`,
where `_group_notes()` buckets a chord's symbols by duration and the bucket
index becomes `rhythmic_layer` — an index local to one chord, which I take to be
why `rebalance_measure_voices()` reassigns every `<voice>` per measure
afterwards.

So that level can only key on the staff, and keying on staff rather than voice
takes the pass from 67 ties to 80 on my test material. BWV 846's second page
alone goes from 8 to 21, and all 14 it adds pair notes in *different* voices on
the bass staff — a held pedal note against a moving line.

Left it in `build_part` for that reason.

aicelen:
Hi,
thank you for the PR :)
Is it expected that it doesn't work between two chords?

Input image:
test_141_2


Rendered with musescore (still marked as slur):
image

The same thing happens in measure 4 and 5 of tabi.jpg

weixlu:
I also noticed it doesn’t work on chords. Since it flattens all notes (`for note in measure.findall("note")`), chord information isn’t handled. That’s why I previously suggested putting tie detection on `list[SymbolChord]` — it would make chords easier to deal with. But @pschuegr said he has tried this and find voice info unavailable. 

So it seems supporting chords may be hard.

pschuegr:
Aha, nice find - I should have tested this.  Let me look at it.

pschuegr:
@aicelen good catch, and yes it was broken — thanks for the image, it made this
easy to reproduce.

The pass walked notes one at a time, so a chord's own members were each other's
neighbours: for C-E-G → C-E-G the pairs it considered were C→E, E→G, G→C, none
of which share a pitch, and no tie between two chords was ever found. Fixed in
`565f7d7`.

Adjacency is now the *event*, a note together with its `<chord>` members. A
single note is an event of one, so the previous behaviour is a special case
rather than a second path.

Checked against `figures/tabi.musicxml`, whose measures 4 and 5 hold the same
chord on the lower staff — before, no tie and the slur kept; after, B3 tied. And
against the encodings two of my test scores were engraved from:

| | truth | found | correct | wrong | precision | recall |
|---|---|---|---|---|---|---|
| BWV 846, before | 68 | 65 | 63 | 2 | 0.97 | 0.93 |
| BWV 846, after | 68 | 65 | 63 | 2 | 0.97 | 0.93 |
| Beach op. 44, before | 40 | 13 | 13 | 0 | 1.00 | 0.33 |
| Beach op. 44, after | 40 | 18 | 18 | 0 | 1.00 | 0.45 |

BWV 846 is unchanged, having no tied chords, so the fix has nothing to do there.
All five ties it adds on the Beach are in the source.

Tests are at 15: two chords tied, one notehead of a chord tied, a chord slurred
to a different pitch staying a slur, and two guards for the conditions the rule
rests on — a curve that leaves a pitch and returns to it an event later stays a
phrase, and a pitch shared by two chords stays untied when the curve is not on
it.

@weixlu you were right that chords were the weak spot. It turned out not to
need `SymbolChord` though: the event grouping is recoverable from the XML in a
couple of lines, so the pass can still key on the voice.

weixlu:
Thanks! I run a score on your latest code, and it works, the slur on two chords is recognized.

aicelen:
@weixlu I think we can merge this now

weixlu:
OK. @pschuegr thanks for your patch, I have merged it.

PR #143, opened 2026-08-25
support AMD GPU in inference

weixlu:
This patch enables inference on AMD GPU. It is fairly straightforward: ONNX already supports AMD, so we just need to turn it on.

There are two patches:

1. **support AMD GPU in inference** — core implementation, covering both `segment` and `transformer`. Per ONNX’s requirements, AMD GPUs still use `cuda` as the device name, so most of the code can be reused as-is.

2. **dockerfile refactor and add support for AMD GPU** — adds a Dockerfile for the AMD GPU environment (I have build it; it works). We now have three Dockerfiles, so I grouped them into a single directory.

Again the code itself is simple. But 3 things here worth noting:

0. **Poetry install.** Before this patch, run `poetry install` or `uvx homr ` is enough to run homr on CPU. But due to compatibility issue, now we have to specify a inference backend (CPU, NV GPU or AMD GPU)

1. **Python version.** 3.11 is recommended, but I have been using 3.12 without issues. The AMD GPU environment only supports 3.10 and 3.12, so I chose 3.12 in the Dockerfile.

2. **ROCm version.** ROCm is AMD’s driver stack (like CUDA on NVIDIA, but less stable than CUDA). This patch is based on ROCm 7.0, and it shall be OK with earlier version (like 6.0). The latest ROCm 7.2 (released late last year) changed API fundamentally, so it can't run.


liebharc:
Hi, from the code it looks good. I haven't run any tests with it as I unfortunately don't have the time right now. Feel free to merge it and thanks!

PR #144, opened 2026-08-27
Preload pip-wheel CUDA libraries so the CUDA provider actually loads

knoguchi:
With `onnxruntime-gpu[cuda,cudnn]`, CUDA/cuDNN live in `nvidia-*` pip wheels that the dynamic loader doesn't find. `CUDAExecutionProvider` is reported as available, but session creation fails to load it and homr silently falls back to the CPU (~30 min instead of ~6 s per page on an RTX 4070 Ti Super).

Calling `onnxruntime.preload_dlls()` at import makes onnxruntime load the wheel libraries itself. Guarded for onnxruntime < 1.21 and CPU-only hosts.


aicelen:
Hi,
Thank you for the PR. 
How does this PR change the existing behaviour? We already have ort.preload_dlls() in line 33 homr/segmentation/inference_segnet.py.

weixlu:
@aicelen pointed out that `ort.preload_dlls()` is already called in Segnet's initialization. However, Segnet is sometimes skipped (see `loaded_from_cache` in `extract()`) and transformer runs directly — in that case `ort.preload_dlls()` never gets executed. So it makes sense to move `ort.preload_dlls()` out of Segnet into a common place, like this PR. @knoguchi, would you mind also removing `ort.preload_dlls()` from `Segnet.__init__()` in this PR?

PR #145, opened 2026-08-27
Decide tuplet removal by beat checksum instead of below-median duration

knoguchi:
The below-median heuristic in `_fix_over_eager_tuplets` strips correct tuplets en masse on uniformly tuplet-heavy pieces (Moonlight Sonata mvt 1: every accompaniment measure is triplets, so ~half sit at/below the median and lose them, and the mangled durations then corrupt note placement in the MusicXML generator).

This anchors the expected measure duration on tuplet-free measures and strips a measure's tuplets only when stripping lands it on that expectation while keeping them misses it; in doubt it preserves what the model read.

Measured on an IMSLP scan of Beethoven op.27/2 (page 2, 19 measures) against the kern ground truth: mean pitch F1 0.951 → 0.996, 18/19 measures pitch-perfect.


knoguchi:
I didn't notice the main branch has some improvements.  the main already achieved 0.996.

PR #147, opened 2026-08-29
Use data from segementation model

aicelen:
This PR introduces a new dataset using the outputs of homr's first pipeline. The aim is to give the model data similar to what it gets during inference. 
The dataset I used is this one: huggingface.co/datasets/guangyangmusic/OpenScore-StringQuartets (contains scores from IMSLP, so this is not synthetic data).
I know that we already use the dataset in the Lieder pipeline. However, this approach would allow us to rule out faulty data as a potential cause of the poorer model performance.

How it works:
1. Put scores and the corresponding musicxml into the sql database
2. Run homr on them and save the generated staffs and additional information to the database
3. Create token files using the the original musicxml file and layout information from homr

It currently generates around 4000 image-token pairs which I used for fine tuning without seeing any improvement in homr's performance. I assume that this is primarily due to the small dataset size.

Things I'd like to try:
- finetuning only the decoder (not both)
- using a larger dataset
- augment the score images before running them through homr


homr/segmentation/inference_segnet.py
```diff
@@ -195,6 +195,7 @@ def inference(
 
     global _segnet_inference  # noqa: PLW0603
     if _segnet_inference is None:
+        print(f"LOADING SEGNET, GPU: {use_gpu_inference}")
```
liebharc: Use eprint

homr/main.py
```diff
@@ -116,8 +131,10 @@ def load_and_preprocess_predictions(
     image = autocrop(image)
     image = resize_image(image)
     preprocessed = color_adjust.apply_clahe(image)
-    predictions = get_predictions(image, preprocessed, image_path, enable_cache, segnet_use_gpu)
-    debug = Debug(predictions.original, image_path, enable_debug)
+    predictions = get_predictions(image, preprocessed, image_path, enable_cache, segnet_use_gpu, data_gen != -1)
+    debug = Debug(predictions.original, image_path, enable_debug, data_gen != -1)
+    if debug.data_gen:
```
liebharc: I haven't thought long about this but at first it looks convoluted: data_gen gets assigned to debug only to be checked so that then the db_page_index gets set. 

Also, the page index conceptionally doesn't seem to be a debug flag. 

Can we possibly just remove data_gen and db_page_index from debug again? Again I only thought about this briefly but it looks like it doesn't need to be there. main already passes the flag to where it is needed.

training/omr_datasets/create_data_from_sq.py
```diff
@@ -0,0 +1,99 @@
...
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+from datasets import load_dataset
+
+from homr.sql_database import datagen_db, org_images_path
+from homr.simple_logging import eprint
+from validation.tools import _REPO_ROOT
+
+
+def run_homr(image_path: Path, img_index: int) -> None:
```
liebharc: What is the sq in the script name (create_data_from_sq)? I first thought it's sql but after reading the script, I'm not sure anymore.

training/omr_datasets/music_xml_parser.py
```diff
@@ -553,12 +552,12 @@ def _collect_articulation(note: ET.Element, part: TokensPart, staff: int) -> tup
         return empty, empty
 
     if len(articulations) == 0:
-        return empty, str.join("_", sorted(slurs))
+        return empty, str.join("_", sorted(set(slurs)))
```
liebharc: looks like an unrelated improvement, but a good one

training/transformer/data_loader.py
```diff
@@ -73,8 +73,8 @@ def __getitem__(self, idx: int) -> Any:
             random.seed(idx)
             np.random.seed(idx)
 
-        img = distort_image(img, allow_occlusions=not self.is_validation)
-        img = add_image_into_tr_omr_canvas(img)
+        # img = distort_image(img, allow_occlusions=not self.is_validation)
```
liebharc: left over debug code?

training/transformer/train.py
```diff
@@ -146,13 +147,9 @@ def train_transformer(
             shutil.rmtree(os.path.join(git_root, checkpoint_folder))
 
     dataset_index = [
-        lieder_train_index,
-        grandstaff_train_index,
-        primus_train_index,
-        pdmx_train_index,
-        musetrainer_train_index,
+        datagen_train_index,
```
liebharc: that should likely be an option in the script: Full training vs fine tuning

homr/staff_parsing.py
```diff
@@ -308,28 +311,39 @@ def parse_staff_image(
...
-            cv2.putText(
-                result_image,
-                str(i) + ": " + symbol.rhythm,
-                (center_int[0], center_int[1] - 10),
-                cv2.FONT_HERSHEY_SIMPLEX,
-                0.3,
-                (0, 0, 255),
-                1,
-            )
+    if debug.data_gen:
```
liebharc: I see, here debug.data_gen is needed. An alternative would be to convert the staff parsing into a class and then set an instance flag on it. However, my idea might be overkill.

aicelen:
Closing this for now, since the visual difference between data generated and lieder generated data is really small when using synthetic sources.

Issue #148, opened 2026-08-30
Homr is slow on big pages


aicelen:
I tried running a 30+ page long pdf through homr, but the staff extraction (after running segmentation) is really slow. This is caused by the larger image size (currently the pdf render just stacks the image vertically).

We could optimize the processing step or just render the pdf to individual images, running homr on them and then using relieur to merge the generate musicxmls into one file. 

To be honest I think the second idea is better, at the same time we could add support for multiple images. What do you think?

weixlu:
Yes, the second option sounds a lot easier to implement. I've never used relieur though, so I don't know how good its merging results are.

Adding support for multiple images also makes a lot of sense to me — most scores span more than one page anyway

Issue #5, opened 2024-06-04
How to run on multiple pages?


mm3509:
I have sheet music on multiple pages. I could append the several files horizontally or vertically, but doing so over multiple files is tedious. Could HOMR provide a way to combine or concatenate the several pages at the output instead of the input?

liebharc:
Hello @emonigma , just to double-check: What you would need is that homr takes e.g. image1.jpg, image2.jpg, image3.jpg and gives you a single musicxml file with the contents of all three images?

emonigma:
Yes, exactly. I could learn about the XML format of MusicXML, or how to concatenate them in MuseScore, but I think it would be a nice use case of getting several images into a single XML file.

liebharc:
I was thinking about this: `homr` detection isn't perfect and likely never will be. The transformer at its core has some randomness, which can yield slightly different results for the same input. So, if you have several pages of sheet music, a good workflow might be:

- Process each file individually, repeating the processing if the results are poor.
- Concatenate the resulting musicxml files.

This approach allows you to select the best result for each page, rather than hoping the processing of all files gives you a good result.

From this perspective, a separate script to concatenate musicxml files might be a better solution. I might add one in the future if I find the time—right now, my focus is still on improving detection in `homr`. However, if you'd like to contribute such a script, I'd be happy to accept a PR!

The script itself shouldn't be hard to write:

- Take everything from the first file.
- Open the second file and, for each part node, copy all measure nodes and append them to the corresponding part of the first file.
- The clef, key, and time signatures will appear multiple times (once for each page), so you'll likely want to remove the duplicates unless there's a change in keys.

There is also this file https://github.com/liebharc/homr/blob/main/training/music_xml.py which parses musicxml files during training. However, I regret using xmltodict for it. For xml generation in `homr` (https://github.com/liebharc/homr/blob/main/homr/xml_generator.py) I switched to `musicxml` which works much better. Both scripts might help in writing a concatenation script.

emonigma:
Hi! That's a good idea and I tested it on a melody that had two pages, then joined them in MuseScore. It works! But fixing the piano chords to coincide with the original ended up being a lot of work, so for my use case, I decided to simply hire a musician to record my scores. I'm sorry I won't be using HOMR in the immediate future, but it's good to know that this technology exists and provides good results in case I want to publish these scores.

Feel free to close the issue. Thanks!

liebharc:
[...]

PR #149, opened 2026-08-31
Multi page support and optimize pdf processing

aicelen:
This introduces support for multiple pages by using relieur to merge generated musicxmls.
It also fixes the issue #148 by rendering the pdf on individual images instead of one large image.


aicelen:
Seems like musicxml is not in our poetry env. It only supports to python <3.13 and it's last commit and release was end of 2024. What do you think should we do?

aicelen:
> Seems like musicxml is not in our poetry env

I decided to replace it with the xml libary.

weixlu:
The PR came together so quickly—you’re very efficient. I looked through the code, and it is clear and easy to follow. I don’t see any issues with it.

However, I’d like to discuss homr’s input and output behavior:

1. Before this PR, when the input was a directory, homr generated one MusicXML file for each image inside it—i.e. without merging them. Validation code also relies on this behavior.
2. In this PR, all images in a dir are combined into a single MusicXML file.

The input/output behavior is an important user-facing interface, so I think we should discuss it carefully. We could also ping @liebharc 

I think merging all images in a dir could cause issues, because multi-page support depends heavily on the page order—i.e. which image is the first page, which is the second. In the code, we use `sorted()` to determine the order of images in dir. However, users may not understand how Python sorting works, especially across different OS or when filenames are not plain English.

I'd suggest this to be input pattern
```bash
# 1. Single image: an independent score
homr page.png

# 2. Multiple images: consecutive pages of the same score, merged in the given order
homr page-01.png page-02.png page-03.png

# Independent images in a directory: process each one separately, without merging
homr scans/
```

homr/main.py
```diff
@@ -367,6 +366,31 @@ def download_weights(segnet_use_gpu: bool, transformer_use_gpu: bool, coreml_enc
...
+        try:
+            xml_paths.append(process_image(image_file, config, xml_generator_args))
+            eprint("Finished", image_file)
+        except Exception as e:
+            eprint(f"An error occurred while processing {image_file}: {e}")
+            error_files.append(image_file)
+    if len(error_files) > 0:
+        eprint("Errors occurred while processing the following files:", error_files)
+
+    if len(xml_paths) > 1:
```
weixlu: - when processing multiple images without merging, it's OK to skip error files. 
- when merging is requested, the entire score should fail if recognition fails for any page.

aicelen:
I think your approach is a good but let's wait what liebharc thinks. 

Edit: It would also be better to run homr when using a pdf file on a list of paths instead of all files in the generated_pdf directory.

liebharc:
@weixlu , I like your proposal, here just my notes:

```
# 1. Single image: straight-forward
homr page.png

# 2. Multiple images: good addition
homr page-01.png page-02.png page-03.png

# Independent images in a directory: that's the current behavior and yes I'd like to keep it. I use that to run homr on a set of independent files which I collect in a folder
homr scans/
```

Also, if you add [relieur ](https://github.com/papoteur-mga/relieur) then it would be nice to let https://github.com/papoteur-mga know (perhaps open an issue in the relieur repo to inform them). Back in https://github.com/liebharc/homr/issues/5 I discussed the different options with him, and it was actually my suggestion to have the merge code outside homr. In hindsight that was possible not the best of my ideas :).

papoteur-mga:
Hello,
I have no problem to see relieur integrated here.
Good job.
The important point is to not have issue in ordering the pages.

aicelen:
I also added merging for pdf+pdf and pdf+image. How do we want to handle pdf files in directories (currently this will give an error)?

@papoteur-mga Thank you! I switched from musicxml to ET.element, should i open a PR in relieur as well?

papoteur-mga:
> @papoteur-mga Thank you! I switched from musicxml to ET.element, should i open a PR in relieur as well?

I see that. No, this is not needed, I will close relieur as it will have no more interest now that the code is inside homr. I don't know of any usage outside homr.

weixlu:
I have read the code and it looks good! As for handle pdf files in directories, I'm not sure 😂 I think it's ok if we render it into multi-pages, or just leave it as unsuppoerted

aicelen:
I decided to merge each pdf into one musicxml. I'll merge the PR this evening.

Issue #152, opened 2026-09-03
PDMX dataset low quality examples


aicelen:
As we already discussed, the model 426 is only a little bit better than model 414 even tough it was trained on two new datasets.
A possible reason for that could be low resolution examples making it hard for the model to give correct outputs without memorization. 

Example 1:
<img width="1280" height="256" alt="Image" src="https://github.com/user-attachments/assets/1df55328-f981-47ac-8353-1b7cd07fe332" />

Example 2:
<img width="1280" height="256" alt="Image" src="https://github.com/user-attachments/assets/094228f5-3cb6-4a52-ab00-4f026fd92402" />

I tested Example 2 both with model 414 and 426. While 426 gives pretty good outputs, 414 doesn't. 

I also tested the first staff of tabi.jpg downsampled (300 -> 96 -> 256). In this case 426 performs better than 414. This means that some generalization happens and the model therefore performs better on low resolution examples. I think it would still be good to filter low resolution staffs because the event of having a low resolution input during inference is quite unlikely (phone cameras usually produce images with 12MP).

The question is how we filter the examples. My ideas:
- Run the segnet on it and filter by average notehead size
- Filter by sequence length (easiest but has obvious flaws)
- Filter by ratio of the staff (would need to distinguish between single and grandstaff images)
- Distance between the staves

What do you think?

Edit: I'd guess that around 5% of images are affected. Apart from that there's also an issue with corrupted slurs like this:
<img width="1280" height="256" alt="Image" src="https://github.com/user-attachments/assets/19c8f40c-26c2-44dd-bf2f-3c5a6d17677a" />

weixlu:
I’ve also found that low-resolution (blurry) images often can't be recognized accurately. However, I think this should be left to users, who need to provide a clear score image. #153 looks like a simple implementation (compared with filtering by average notehead size).

liebharc:
> Filter by ratio of the staff (would need to distinguish between single and grandstaff images)

That would be my first thought. Find examples which are very wide (before we place them on the TrOMR canvas and thus normalize the width) and skip them. If in doubt in might be better to filter out too many examples, then keeping bad ones. We have plenty of data I believe, and we already found in the past that only a few bad examples are needed to degrade performance.

PR #153, opened 2026-09-03
Filter files with wrong aspect ratio

aicelen:
See #152 for a more detailed description.

MAX_RATIO describes the max value for width/height per grandstaff. 
For the current value of 8 this means: 1280/8 = 160. So a rendered grandstaff needs to have at least 160 px in height.

I haven't trained yet.

### DIFF
--- a/homr/transformer/staff2score.py
+++ b/homr/transformer/staff2score.py
@@ -87,7 +87,8 @@ def test_transformer_on_image(path_to_img: str) -> None:
     """
 
     model = Staff2Score(Config())
-    image = Image.open(path_to_img)
+    image = np.array(Image.open(path_to_img))
+    image = np.dot(image[..., :3], [0.2989, 0.5870, 0.1140])
     out = model.predict(np.array(image))
     eprint(out)
 
--- a/training/omr_datasets/convert_musetrainer.py
+++ b/training/omr_datasets/convert_musetrainer.py
@@ -6,6 +6,7 @@ import random
 import signal
 import subprocess
 import sys
+import uuid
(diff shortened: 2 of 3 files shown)


weixlu:
How about I run the training for this PR

aicelen:
Yes, that would be great :)
It filtered 5500 staffs so we might need to increase the value for _TARGET_FILES

aicelen:
@weixlu Any updates on this?

weixlu:
Ah yes, I set `_TARGET_FILES = 12000` and trained 33 epochs. SMB scores: 14.50% and Polish: 17.87%, looks similar to previous one. I'll send you the model later

aicelen:
Thank you for the training run. I'm really surprised the model didn't improve more. Although it shows some improvements on tabi.jpg it is still not as good as 414. 
I have two ideas:
- 426 was only trained for 27 epochs, if you still have a checkpoint we could benchmark that as well
- Maybe the filtering was not aggressive enough, I'll do some more testing (I’ll have time for that this weekend)

weixlu:
> 426 was only trained for 27 epochs, if you still have a checkpoint we could benchmark that as well

If I get it right, this means resume training 426 from epoch 27 to 35. But 426 was trained by liebharc (414 was trained by me), so I’m afraid I can’t resume it. I could start a new training from scratch if you need it.

aicelen:
Oh, I didn't express myself clearly. I meant to test model 448 at epoch 27

weixlu:
OK, I got it, here is the result: smb 14.66%, polish 18.40%. slightly worse than a full training.

PR #154, opened 2026-09-04
Say what a count mismatch actually is

eerovil:
Across the 93 printed systems of this repertoire the check reported **505** moments holding a different number of noteheads. That reads as five hundred misread notes. Checked one at a time, **469 of them are not notes homr got wrong** — and the two causes behind them want opposite work.

## What the 505 are

**328 sit in nineteen systems that disagree about how many staves are printed.** Seventeen of the nineteen write three or four staves where the page prints two, having given a divisi staff's second voice a staff of its own. The staff is part of the key every note is matched on, so past the first staff that diverges the two files are no longer being compared on the same music. 54 of the 128 wrong pitches are in there too.

The clearest case is `kaksi-laulua-krapulasta-2-s13`, which reported **26 lost noteheads on a system homr read note for note correctly**:

```
REF : P1 v1 14  D4 E4 D4 D4 C4 E4     HOMR: P1 v1 14  D5 E5 D5 D5 C5 E5
      P1 v2 14  B3 B3 B3 B3 A3 C4           P1 v2 14  B4 B4 B4 B4 A4 C5
      P2 v1 22  D3 D3 D3 D3 E3 F3           P2 v1 22  D3 D3 D3 D3 E3 F3
      P2 v2 28  D3 D3 C3 C3 B2 A2           P3      28  (the same line)
```

All 78 notes present, all right, on three staves instead of two. (The octave is the male-choir writing the comparison already handles.)

**141 of the remaining 177 are the bar's own rhythm.** homr wrote nothing at that beat while holding that bar's notes at other beats — a duration read differently early in the bar walks the cursor out of step, and every moment after it reports one against zero. **All 141 are that shape; not one is a bar homr left empty.**

**36 are left.** The moment both files agree exists and disagree about: two heads printed and one written, or one printed and two written. That is the number worth chasing.

## The change

`structure` is checked first and counted **once per case**, not once per note, and the case page opens by saying the rows beneath it are its consequence. `timing` is separated from `size`. Nothing is discarded and no row is hidden — the counts are split so the number a reader sees means what it says.

| | was | now |
|---|---|---|
| count mismatches | 505 | 36 `size` + 232 `timing` + 19 structural cases |

## What this says about the work

The largest single lever on these numbers is not per-note detection. Nineteen of 93 systems get the staff count wrong, and that one answer is behind 328 of the count mismatches and 42% of the wrong pitches.

## Checked

Harness only — no file under `homr/` is touched, so no parse changes. Recounted all 93 cached parses at `ef4620e`. `pytest tests/` 66 passed; the one failure (`test_poetry_config::test_isset_precommit_hooks`) and the 7 uncollectable modules (`torch`, `editdistance`) are pre-existing and environmental — this is a git worktree, so there is no `.git/hooks` directory.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


eerovil:
Opened by mistake against upstream — this belongs on the eerovil fork. Sorry for the noise.

PR #156, opened 2026-09-07
parse multi voice in staff

weixlu:
## Brief introduction

As noted in #126, it's hard for homr to recognize multiple voices on a single staff, for example:

image

That is because the current vocabulary only distinguishes `upper` and `lower`. This PR further distinguishes `upper`/`upper2` and `lower`/`lower2`. (I used to think this would not work — for example, what do 1 and 2 actually mean? Which voice comes first?) After training, this approach looks viable. Now homr’s recognition results look like this:

image

## Code details

This PR has two parts.

### Kern parser refactor

The existing parser is complex and hard to follow, especially in these two places:

1. It first merges multiple voices on the same staff, then parses. That is unnecessary; we can parse and merge in a single pass. This one-pass behaviour also helps this PR.
2. In smb dataset, one kern file can contain multiple documents, so the original parser logic was not reused. A separate `_parse_kern_document` was added instead. But I found that the logic can actually be reused.

I compared dataset parsing before and after the refactor:

1. The `.tokens` produced for grandstaff dataset are unchanged, except that about 100 low-quality files are now filtered out. I am confident this refactor will not affect training.
2. The smb dataset is more complicated, with many corner cases. After the refactor, NED changes slightly. I compared the diffs with my best effort; I believe those differences come from gaps in the original parser, i.e. the refactored code is closer to ground truth.

### Kern / MusicXML parser changes

This part is more straightforward. MusicXML already has voice data; for kern, voice data can be read from parallel spines. 

I retrained the model. Benchmark results are polish 16.89% and smb 13.72%, so there appears to be no regression at least.

This PR is still fairly complex; we can take our time reviewing it. And there are quite some test code, which are AI-generated, while other code is hand written by me

### DIFF
--- a/homr/transformer/vocabulary.py
+++ b/homr/transformer/vocabulary.py
@@ -94,8 +94,12 @@ def build_position() -> dict[str, int]:
     """
     The staff position, applies to notes, rests and clefs
     """
-    positions = [nonote, "upper", "lower"]
+    positions = [nonote, "upper", "upper2", "lower", "lower2"]
     return build_dict(positions)
+
+
+def is_lower_position(position: str) -> bool:
+    return position.startswith("lower")
 
 
 def build_articulation() -> dict[str, int]:
@@ -337,10 +341,10 @@ class EncodedSymbol:
         return result
 
     def to_upper_position(self) -> "EncodedSymbol":
-        if self.position != "lower":
+        if not is_lower_position(self.position):
             return self
         result = copy.copy(self)
-        result.position = "upper"
+        result.position = self.position.replace("lower", "upper")
         return result
 
     def is_valid(self) -> bool:
@@ -472,7 +476,7 @@ def _remove_redudant_clefs_keys_and_time_signatures(
         result = []
         for symbol in chord:
             if symbol.rhythm.startswith("clef"):
-                if symbol.position == "upper":
+                if not is_lower_position(symbol.position):
                     if symbol.rhythm != clef_upper:
                         clef_upper = symbol.rhythm
                         result.append(symbol)
@@ -607,7 +611,7 @@ def _only_keep_lower_staff_if_there_is_a_clef(
         for symbol in chord:
             if has_lower_clef:
                 result.append(symbol)
-            elif i < 5 and symbol.rhythm.startswith("clef") and symbol.position == "lower":
+            elif i < 5 and symbol.rhythm.startswith("clef") and is_lower_position(symbol.position):
                 has_lower_clef = True
                 result.append(symbol)
             else:
--- a/training/transformer/training_vocabulary.py
+++ b/training/transformer/training_vocabulary.py
@@ -10,6 +10,7 @@ from homr.transformer.vocabulary import (
     Vocabulary,
     empty,
     has_rhythm_symbol_a_position,
+    is_lower_position,
     nonote,
     sort_token_chords,
 )
@@ -61,7 +62,7 @@ def check_token_lines(lines: list[EncodedSymbol]) -> None:
 
 
 def _symbol_to_sortable(symbol: EncodedSymbol) -> int:
-    position = 10000000 if symbol.position == "lower" else 0
+    position = 10000000 if is_lower_position(symbol.position) else 0
     if "note" in symbol.rhythm:
         return (
             vocab.pitch[symbol.pitch] * len(vocab.rhythm) + vocab.rhythm[symbol.rhythm] + position
@@ -83,12 +84,12 @@ def _chord_to_str(chord: list[EncodedSymbol]) -> str:
         artic_stripped, symbol_stripped = symbol.strip_articulations([], remove_all=True)
         slur_stripped, symbol_stripped = symbol_stripped.strip_slurs([], remove_all=True)
         for articulation in artic_stripped:
-            if symbol.position == "lower":
+            if is_lower_position(symbol.position):
                 lower_artics.add(articulation)
             else:
                 upper_artics.add(articulation)
         for slur in slur_stripped:
-            if symbol.position == "lower":
+            if is_lower_position(symbol.position):
                 lower_slurs.add(slur)
             else:
                 upper_slurs.add(slur)
@@ -107,7 +108,8 @@ def _chord_to_str(chord: list[EncodedSymbol]) -> str:
 
     if len(upper_slurs) > 0:
         first_upper = next(
-            (idx for idx, s in enumerate(annotation_resorted) if s.position != "lower"), None
+            (idx for idx, s in enumerate(annotation_resorted) if not is_lower_position(s.position)),
+            None,
         )
         if first_upper is not None:
             annotation_resorted[first_upper] = annotation_resorted[first_upper].add_slurs(
@@ -115,7 +117,8 @@ def _chord_to_str(chord: list[EncodedSymbol]) -> str:
             )
     if len(lower_slurs) > 0:
         first_lower = next(
-            (idx for idx, s in enumerate(annotation_resorted) if s.position == "lower"), None
+            (idx for idx, s in enumerate(annotation_resorted) if is_lower_position(s.position)),
+            None,
         )
         if first_lower is not None:
             annotation_resorted[first_lower] = annotation_resorted[first_lower].add_slurs(
@@ -124,7 +127,8 @@ def _chord_to_str(chord: list[EncodedSymbol]) -> str:
 
     if len(upper_artics) > 0:
         first_upper = next(
-            (idx for idx, s in enumerate(annotation_resorted) if s.position != "lower"), None
+            (idx for idx, s in enumerate(annotation_resorted) if not is_lower_position(s.position)),
+            None,
         )
         if first_upper is not None:
             annotation_resorted[first_upper] = annotation_resorted[first_upper].add_articulations(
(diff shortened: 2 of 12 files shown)


liebharc:
I hadn't time to look into the PR yet, and it might take a while until I find the time, sorry. One thing from the description: The transformer needs clear rules to learn. No ambiguities. What is the rule to find out if something is upper or upper2?

A visual clue is the stem direction, just that our ground truth data doesn't have this available.

weixlu:
Sure, no hurry. 

> What is the rule to find out if something is upper or upper2?

Hoestly there hasn't been a clear rule (like stem direction), just parses the kern/musicxml file from begin to end, and who comes first, who belongs to upper1. But the result of this approch seems OK. I'll test more scores later and see if it all works well

weixlu:
I'm not able to find many scores for this case, I just find 3, and all look good on the new model. And it seems the model would prefer to place `upper2` on lower notes

homr/transformer/vocabulary.py
```diff
@@ -337,10 +341,10 @@ def change_lift(self, lift: str) -> "EncodedSymbol":
         return result
 
     def to_upper_position(self) -> "EncodedSymbol":
-        if self.position != "lower":
+        if not is_lower_position(self.position):
```
liebharc: Just minor: For readability we could also introduce is_upper_or_has_no_position, so that we can make it clearer what not lower means. Even better if you find a better function name than my proposal, but still makes it clear that we are dealing with 3 principal positions (upper, lower, none) - well and now upper and lower have to variants, but I don't think that needs to be in the function name.

training/transformer/training_vocabulary.py
```diff
@@ -173,12 +178,13 @@ def max_ledger_lines(tokens: list[EncodedSymbol]) -> int:
     }
     max_ledger_lines = 0
     for symbol in tokens:
+        family = symbol.position.removesuffix("2")
```
liebharc: I think I understand the reason behind "family". It's just still not intuitive. My first thoughts would be: Still position, or direction. With a preference of "position" or "position_type" to not introduce a completely new term here.

liebharc (commented):
I finally found time to look into the code :). Looks good overall, although also I didn't look into everything in detail. 

Most of all I'm still pondering the question of if the rule for upper vs upper2 (and same for lower) is a good rule in the sense of: Can the transformer learn it. I think there are two angles here:

- We could try to improve the rule, not sure how yet
- Or we just accept that it works well enough apparently as your training run seems to show

I'm not fully decided on this question yet.

liebharc:
I checked the data and I think I now better understand the idea. And yes this should work as it considers the stem direction.

The AI Overlord claims that this diff on top of your branch would help. And working through the data it seems to actually improve. An example is:  lc4946175 measure 15

So overall I'd say lets add this to your PR and then feel free to merge! I'm curious to see how it works out!

```
diff --git a/tests/test_music_xml_parser.py b/tests/test_music_xml_parser.py
index 6631f3d..aefacb8 100644
--- a/tests/test_music_xml_parser.py
+++ b/tests/test_music_xml_parser.py
@@ -561,6 +561,43 @@ note_2 D4 _ _ _ upper
 barline . . . . ."""
         self.assertEqual(token_str, expected)
 
+    def test_stem_up_voice_is_main_voice(self) -> None:
+        """If the first voice has stems down and exactly one other voice has stems up,
+        then the stem up voice is the main voice. Otherwise the first voice stays the
+        main voice."""
+        example = """<score-partwise version="4.0"><part id="P1">
+<measure number="1">
+  <attributes>
+    <divisions>1</divisions>
+    <clef number="1"><sign>G</sign><line>2</line></clef>
+  </attributes>
+  <note>
+    <pitch><step>C</step><octave>4</octave></pitch>
+    <duration>1</duration><voice>1</voice><type>quarter</type><stem>down</stem>
+  </note>
+  <backup><duration>1</duration></backup>
+  <note>
+    <pitch><step>G</step><octave>4</octave></pitch>
+    <duration>1</duration><voice>2</voice><type>quarter</type><stem>up</stem>
+  </note>
+</measure>
+<measure number="2">
+  <note>
+    <pitch><step>D</step><octave>4</octave></pitch>
+    <duration>1</duration><voice>1</voice><type>quarter</type><stem>down</stem>
+  </note>
+  <backup><duration>1</duration></backup>
+  <note>
+    <pitch><step>A</step><octave>4</octave></pitch>
+    <duration>1</duration><voice>2</voice><type>quarter</type><stem>down</stem>
+  </note>
+</measure>
+</part></score-partwise>"""
+        tokens = music_xml_string_to_tokens(example)
+        symbols = [s for page in tokens for measure in page for s in measure]
+        notes = {s.pitch: s.position for s in symbols if s.rhythm.startswith("note")}
+        self.assertEqual(notes, {"C4": "upper2", "G4": "upper", "D4": "upper", "A4": "upper2"})
+
     def test_voice_assignment_is_per_staff_and_measure(self) -> None:
         """Hidden rests do not claim a voice; each staff and measure starts fresh."""
         example = """<score-partwise version="4.0"><part id="P1">
diff --git a/training/omr_datasets/music_xml_parser.py b/training/omr_datasets/music_xml_parser.py
index 214528f..145462f 100644
--- a/training/omr_datasets/music_xml_parser.py
+++ b/training/omr_datasets/music_xml_parser.py
@@ -1,4 +1,5 @@
 import xml.etree.ElementTree as ET
+from collections import defaultdict
 from typing import Iterable, SupportsIndex, TypeVar, overload
 
 from homr.music_xml_generator import DURATION_NAMES
@@ -188,7 +189,8 @@ class TokensMeasure:
         is_grace = "G" in symbol.rhythm
         symbol.position = self._get_staff_position(staff)
         if not invisible:
-            # The first visible voice of each staff is the main voice in this measure.
+            # The main voices are usually set by _find_main_voices before the measure
+            # is processed, otherwise the first visible voice of each staff is used.
             if voice != self.main_voice.setdefault(staff, voice):
                 symbol.position += "2"
         if is_chord:
@@ -426,6 +428,9 @@ class TokensPart:
         self._flush_pending_clefs()
         self._ensure_current_measure().append_position_change(duration)
 
+    def set_main_voices(self, main_voices: dict[int, int]) -> None:
+        self._ensure_current_measure().main_voice = main_voices
+
     def on_end_of_measure(self) -> None:
         self.measures.append(self._ensure_current_measure().complete_measure())
         self.current_measure = TokensMeasure()
@@ -595,6 +600,50 @@ def _collect_articulation(note: ET.Element, part: TokensPart, staff: int) -> tup
     return str.join("_", sorted(articulations)), str.join("_", sorted(slurs))
 
 
+def _is_hidden_note(note: ET.Element) -> bool:
+    return note.get("print-object", None) == "no" or any(
+        _text(note_head) == "none" for note_head in _children(note, "notehead")
+    )
+
+
+def _find_main_voices(measure: ET.Element) -> dict[int, int]:
+    """
+    Find the main voice of each staff in a measure. Notes and rests of the main voice
+    get the position upper/lower, all other voices upper2/lower2.
+
+    The label should follow what is visible in the image: if a staff has several
+    voices, the main voice is drawn with stems up and the others with stems down.
+    Usually that's the first visible voice (which is also the one with the lowest
+    number), but engravers sometimes enter the upper part in the second voice and
+    flip the stems. So if the first voice has mostly down stems and exactly one other
+    voice has mostly up stems, then that voice is the main voice.
+    """
+    first_voice: dict[int, int] = {}
+    stem_score: dict[tuple[int, int], int] = defaultdict(int)  # up stems - down stems
+    for note in _children(measure, "note"):
+        if _is_hidden_note(note):
+            continue
+        staff_nodes = _children(note, "staff")
+        staff = _int_text(staff_nodes[0], 1) - 1 if staff_nodes else 0
+        voice_nodes = _children(note, "voice")
+        voice = _int_text(voice_nodes[0], 0) if voice_nodes else 0
+        first_voice.setdefault(staff, voice)
+        stem = _text(_child(note, "stem"))
+        if stem == "up":
+            stem_score[(staff, voice)] += 1
+        elif stem == "down":
+            stem_score[(staff, voice)] -= 1
+
+    result: dict[int, int] = {}
+    for staff, voice in first_voice.items():
+        up_voices = [v for (s, v), score in stem_score.items() if s == staff and score > 0]
+        if stem_score.get((staff, voice), 0) < 0 and len(up_voices) == 1:
+            result[staff] = up_voices[0]
+        else:
+            result[staff] = voice
+    return result
+
+
 def _process_note(part: TokensPart, note: ET.Element) -> None:
     staff = 0
     note_heads = _children(note, "notehead")
@@ -724,6 +773,7 @@ def _process_multi_rests(part: TokensPart, measure_style: ET.Element) -> None:
 def _music_part_to_tokens(part: ET.Element) -> list[Measure]:
     tokens = TokensPart()
     for measure in _children(part, "measure"):
+        tokens.set_main_voices(_find_main_voices(measure))
         for child in list(measure):
             if child.tag == "attributes":
                 _process_attributes(tokens, child)

```

weixlu:
I think this AI-generated patch is really good. Now `upper`/`upper2` are determined by the stem direction. I've cleaned up the code and I'm about to run training again. If the results look good, I'll merge it.

aicelen:
Does this already use #162 (right now I don't have the time to check myself)?

weixlu:
> Does this already use #162 (right now I don't have the time to check myself)?

Not yet. I notice that your pr also needs a training. I'm not sure if I should include your pr and start one training, or start two trainings for each pr seperately

aicelen:
Okay, I think model 453 is already trained with my PR. Shall I still merge #162 tonight?

Issue #136, opened 2026-08-06
Crash with `ValueError: XMLDuration.value '0' must be greater than '0'`


timber-they:
**Describe the bug**
When converting a specific file (see image below), the program crashes at some point with the error
```
ValueError: XMLDuration.value '0' must be greater than '0'
```

**Input Image**
<img width="1700" height="2200" alt="Image" src="https://github.com/user-attachments/assets/61254fc4-4edf-4cfd-8932-09e9328dcfea" />

**Full Traceback**
```
Traceback (most recent call last):
  File "/home/timber/.cache/uv/archive-v0/k3vB-x2ydk89gsGA/bin/homr", line 12, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/timber/.cache/uv/archive-v0/k3vB-x2ydk89gsGA/lib/python3.13/site-packages/homr/main.py", line 452, in main
    process_image(args.image, config, xml_generator_args)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/timber/.cache/uv/archive-v0/k3vB-x2ydk89gsGA/lib/python3.13/site-packages/homr/main.py", line 211, in process_image
    xml = generate_xml(xml_generator_args, result_staffs, title)
  File "/home/timber/.cache/uv/archive-v0/k3vB-x2ydk89gsGA/lib/python3.13/site-packages/homr/music_xml_generator.py", line 120, in generate_xml
    root.add_child(build_part(args, staff, index, has_two_staves_by_part[index]))
                   ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/timber/.cache/uv/archive-v0/k3vB-x2ydk89gsGA/lib/python3.13/site-packages/homr/music_xml_generator.py", line 134, in build_part
    measures = build_measures(args, voice, is_first_part, has_two_staves)
  File "/home/timber/.cache/uv/archive-v0/k3vB-x2ydk89gsGA/lib/python3.13/site-packages/homr/music_xml_generator.py", line 181, in build_measures
    for note_xml in build_note_chord(staff_pos, state, chord_duration):
                    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #157, opened 2026-09-08
Fix zero XMLDuration crash by including all note durations in division calculation

Jimmy-Mendez:
**Summary**

Fixes the root cause of #136 . While aa5c8ce addressed a related rest-merging issue that sidestepped the crash for the specific image in that report, the underlying math bug remains and still triggers on other inputs.

**The problem**

find_division_and_time_signature_nominator computes the LCM of all duration denominators so that fractional durations can be expressed as integers in MusicXML. But it only collects one duration per chord (chord.get_duration(), which returns the minimum), so individual notes within a chord that have different denominators are never factored into the LCM.

When a note's fraction denominator isn't represented in the LCM, int(fraction * division) truncates to 0 — for example, a tuplet note with fraction = 1/7 in a chord whose min duration is 1/4 gives int(1/7 * 4) = 0. MusicXML rejects duration 0 with ValueError: XMLDuration.value '0' must be greater than '0'.

**The fix**

1. Root cause (find_division_and_time_signature_nominator): Iterate all individual note/rest symbols in each chord when collecting durations for the LCM, not just the chord-level minimum. This ensures every denominator is covered.
2. Safety clamp (build_note_or_rest, build_backup): Wrap duration calculations with max(1, int(...)) as a belt-and-suspenders guard against any remaining edge cases that could produce zero.
3. Bonus bugfix (kern_to_symbol_duration in transformer/vocabulary.py): Added a missing return on the multirest branch (kern.endswith("m")), which previously created the correct SymbolDuration but discarded it and fell through to the general parser.

We hit this on 4 out of 1,000 images from the [Institutional Data Initiative music elements dataset](https://huggingface.co/datasets/institutional/institutional-books-hl-visual-elements/viewer/default/music). All four are vocal+piano scores that homr successfully detects staves for but crashes during MusicXML generation.

All four produce valid MusicXML and render correctly after this fix.

**Files changed:**

- homr/music_xml_generator.py — collect per-symbol durations in division LCM; clamp XMLDuration values to minimum of 1
- homr/transformer/vocabulary.py — add missing return in multirest case


weixlu:
Hi, thanks for the fix! I think it makes sense—the current code indeed doesn't account for cases where a note's duration denominator isn't represented in the LCM.

I'd like to verify this. Could you share a sheet music image that triggers the issue? I can't access the Institutional Data Initiative music elements dataset, as it seems to require permission.

liebharc:
A run on my data showed no regression: 5.1 diffs, SER: 4.0%. So from my side it looks good and it can be merged as soon as @weixlu is fine with it.

Jimmy-Mendez:
Below are a couple of the images that triggered the issue on my end

weixlu:
I have verified the case you mentioned. Thanks for the fix!

PR #158, opened 2026-09-12
Split flat folder into two

aicelen:
I find it annoying that all the lieder training data is in one big folder. Therefore I decided to split it up into a "rendered_scores" (for svg and musicxml files) and a "flat" folder (for png and token files). This also reduces the file size when training in cloud since we don't upload the svg and musicxml files.


weixlu:
looks good :)

liebharc:
Good idea, feel free to merge when you are ready

PR #159, opened 2026-09-13
Experiment/pdmx lieder

aicelen:
I'd like to try training the model without the pdmx pipeline and instead run the data through lieder pipeline. The reason is that I think the pdmx pipeline's data is not representing the actual data well enough (for example beams between eights are always missing and it has some weird case like empty measures).

@liebharc Have you already tried something like this?


training/omr_datasets/convert_lieder.py
```diff
@@ -660,8 +737,13 @@ def convert_lieder(only_recreate_token_files: bool = False) -> None:
         )
 
     eprint("Indexing Lieder dataset, this can up to several hours.")
+
+    from training.omr_datasets.convert_pdmx import _load_filtered_paths
+
+    _create_musicxml_and_svg_files_from_mxl(_load_filtered_paths())
```
weixlu, line 743: It seems PDMX dataset is also being processed inside `convert_lieder.py`, and its outputs are under `datasets/Lieder-main/` as well. It's OK for a quick try, but I’d suggest keeping the PDMX artifacts in their own directory so the two datasets don’t get mixed up.

aicelen:
Yes I'll need to do some clean up before we can merge this (if training turns out to be good). Do you have time and compute to train it (if so keep in mind that the new file isn't integrated into the dataset generation pipeline so you'd need to run it separately)?

weixlu:
Sure, I'll start training. double-check to ensure I get it right: I'll do the following:
1. rm -rf datasets/Lieder-main/
2. poetry run python training/omr_datasets/convert_lieder.py
3. poetry run python training/omr_datasets/filter_from_index.py
4. delete `pdmx_train_index` in `training/transformer/train.py`
5. start training
 is it right?

aicelen:
I'd also try training without musetrainer. Can you access Google drive? If so I can send you a link to the already converted datasets (I already uploaded them there)

weixlu:
> I'd also try training without musetrainer.

So remove `pdmx_train_index` and `musetrainer_train_index` in `training/transformer/train.py`, right?

> Can you access Google drive?

Yes, please send me the link, and I will use yours. btw, I just started convertion, and `QmXoCjhAW2vgtdmSvCS2YMQdTFeZfxXD7wNTSKkMSg1W8v` from pdmx fails (I'm using MuseScore 4.6.5). So i'd suggest add it to `files_with_known_issues`

weixlu:
Found a bug. Take a look at `datasets/Lieder-main/flat/QmdwjkpqGz4HsS6sE168ECrKiLHp67mQDH9X17iPNFNzDh-1-2.png` — it’s extremely narrow, while the original page `datasets/Lieder-main/flat/QmdwjkpqGz4HsS6sE168ECrKiLHp67mQDH9X17iPNFNzDh-1.png` looks fine. For a work around, I remove this file from training dataset. AI suggest this to fix: 
[0001-fix.patch](https://github.com/user-attachments/files/32193256/0001-fix.patch)

liebharc:
> @liebharc Have you already tried something like this?

No, I haven't tried this yet. Good luck! Let's see how it goes :).

aicelen:
Only 10 images are affected by this. If this run turns out to be good, I'd add a small check of image width and height.

weixlu:
Hi, the full training has finished. benchmark shows smb 14.86%, polish 17.76%.

So I guess pdmx dataset may not be as perfect as we expected

aicelen:
This is a slight improvement, right? But this is still worse than I expected. Do you have any other ideas to improve the model?
What do you two think, do we use this in the future as well?

weixlu:
I’m afraid I don’t have a better idea.

I haven’t looked closely at the current `convert_pdmx`, but if we change it to work the same way as `convert_lieder`, the code and later maintenance would look simpler.

aicelen:
@liebharc What do you think, do we remove the pdmx pipeline in favor of this?

liebharc:
You mean in favor for the lieder pipeline? Sure, let's try it. If we would ever need it again, then we could always restore it from version control. So just replace it and see how it goes.

PR #162, opened 2026-09-19
Use Lieder pipeline for pdmx and musetrainer

aicelen:
This is a result of #159, #153 and #152.

In general, the pdmx pipeline produces low quality examples which are not similar to real world examples (for example beams between eights are missing). Therefore it didn't really improve the model and we decided to use the lieder pipeline for pdmx. It should be noted that there's no big improvement on the polish scores benchmark over model 414.
This could have different reasons:
- polish scores ground truth is not good enough (for example image 56 contains a lot of "invisible" tuplets, which the model always gets wrong; also the upper staff and lower staff are getting merged in measure 7)
- the model doesn't scale with more data properly
- the model needs more capacity

It'd be good if one of you could take a look at this PR. After that I'd update to model 453.


weixlu:
Sure, I'll take a look on Monday

training/transformer/train.py
```diff
@@ -123,6 +124,15 @@ def _check_datasets_are_present(selected_datasets: list[str]) -> list[str]:
 
         if dataset == pdmx_train_index and not os.path.exists(pdmx_train_index):
             convert_pdmx()
+    remove_tokens(
+        [
```
weixlu: Here I'd suggest changing this to `remove_tokens(selected_datasets)`. That would be more concise and consistent. It also covers the case where `selected_datasets` contains only grandstaff, so the other datasets have not been downloaded. Running `remove_tokens` on those would then fail because their `index.txt` files cannot be opened.

training/omr_datasets/convert_pdmx.py
```diff
@@ -202,39 +94,37 @@ def convert_pdmx() -> None:
...
         untar_file(mxl_archive, pdmx_root)
 
-    os.makedirs(pdmx_out_root, exist_ok=True)
-
     eprint("Reading CSV and applying filters")
     mxl_paths = _load_filtered_paths()
     eprint(f"{len(mxl_paths)} files pass pre-filters (c<={_MAX_COMPLEXITY}, tracks<={_MAX_TRACKS})")
 
-    with open(pdmx_train_index, "w") as index_f:
+    create_musicxml_and_svg_files_from_mxl(_load_filtered_paths(), rendered_scores)
```
weixlu: Here you could just use `mxl_paths` (see 2 lines above) instead of `_load_filtered_paths()`

training/omr_datasets/convert_pdmx.py
weixlu: Here the same code in `convert_musetrainer.py` can be added
```
    multiprocessing.set_start_method("spawn")
    only_recreate_token_files = False
    if "--only-tokens" in sys.argv:
        only_recreate_token_files = True
```

aicelen:
Can I merge this?

weixlu:
> Can I merge this?

Oh, I see you've push a commit, but I also leave another 2 comments. Could you take a look？

aicelen:
But I can only see one comment :eyes:

weixlu:
Thanks! I think the PR is good to go

training/omr_datasets/convert_lieder.py
weixlu: I'm rebuilding the dataset and found an issue here. 
`makedir` shouldn't be placed at the global level here; otherwise it runs first, and the later `if not os.path.exists(lieder):` is always `False`. I suggest moving it to after this branch.

```
if not os.path.exists(lieder):
    ...

os.makedirs(flat_data, exist_ok=True)
eprint("Indexing Lieder dataset, this can up to several hours.")
```

PR #163, opened 2026-09-20
enhence title detection

weixlu:
This PR improves title detection.

I found that RapidOCR’s results are actually fairly reliable and already support multiple languages, so we only need to pick the text most likely to be the title.

I randomly selected 15 scores with English, German, French, and Chinese titles as test cases. All tests passed. I have not yet decided where to put them; for now they live in a private [Hugging Face dataset](https://huggingface.co/datasets/weixlu/scores)


aicelen:
Didn't test it but the code looks good to me

weixlu:
> Didn't test it but the code looks good to me

Thanks! The code is quite independent (only relates to title detection), so I'll merge it

